PR #21, opened 2019-02-05
CPLAT-4058 Provide additional --help output and better stderr

smaifullerton-wk:
- Created an `ArgParser` with the `--backwards-compat` flag and passed its usage to `runInteractiveCodemod` so that it'll be formatted as an additional `--help` option

- Created `_changesRequiredOutput` and passed into both phases of codemod, so that if either one fails the user will receive a helpful message.

- Fixed a bug, `--help` output is being printed twice, one in each phase of codemod. Now it's only the first one.


aviary-wf:
## Security Insights

## Action Items
---

lib/src/executables/dart2_upgrade.dart
```diff
@@ -61,6 +69,8 @@ void main(List<String> args) {
     ),
     args: args,
     defaultYes: true,
+    additionalHelpOutput: argParser.usage,
+    changesRequiredOutput: _changesRequiredOutput,
   );
 
   if (exitCode > 0) {
```
evanweible-wf: Instead of wrapping the second phase below in an if block, can we add to this existing if block since it already returns early? That should have the same effect and fix that double-help-output bug.

```suggestion
  if (exitCode > 0 || args.contains(_helpFlag) || args.contains('-h')) {
```

lib/src/executables/dart2_upgrade.dart
```diff
@@ -98,20 +108,30 @@ void main(List<String> args) {
...
 
-  exitCode = runInteractiveCodemodSequence(
-    FileQuery.dir(
-      pathFilter: isDartFile,
-      recursive: true,
-    ),
-    [
-      AggregateSuggestor(
-        phaseTwoSuggestors.map((s) => Ignoreable(s)),
+  if (!args.contains(_helpFlag) && !args.contains('-h')) {
```
evanweible-wf: Since we have a constant for the help flag, might as well have one for the abbreviation as well

lib/src/executables/dart2_upgrade.dart
```diff
@@ -36,6 +37,13 @@ import '../ignoreable.dart';
 import '../util.dart';
 
 const _backwardsCompatFlag = '--backwards-compat';
+const _helpFlag = '--help';
+const _changesRequiredOutput = """
+To update your code, switch to Dart 2.1.0 and run the following commands:
+  pub global activate over_react_codemod ^<version>
```
evanweible-wf: Let's put the current version in here to start and I'll update our release event configuration so that it gets updated automatically on every release PR.
```suggestion
  pub global activate over_react_codemod ^1.0.1
```

evanweible-wf:
QA +1
- [x] CI passes
- [x] Verified the stderr output for:
  - `pub run over_react_codemod:dart2_upgrade -h`
  - `pub run over_react_codemod:dart2_upgrade --fail-on-changes`

@Workiva/release-management-p

PR #34, opened 2019-08-05
CPLAT-6811 Update setState to setStateWithUpdater

joebingham-wk:
## Motivation
  
When a consumer is calling `setState` and passing in a function as the first parameter, we want to encourage them to use `setStateWithUpdater` instead. This is a codemod that makes that switch.

## Changes
  
- Add a suggestor that achieves the description above.
- Add test coverage.

## Review

  

Please review: 
@aaronlademann-wf @kealjones-wk @sydneyjodon-wk @greglittlefield-wf 
### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist


aviary-wf:
## Security Insights

## Action Items
---

lib/src/component2_suggestors/setstate_updater.dart
```diff
@@ -0,0 +1,44 @@
...
+
+import '../constants.dart';
+import '../util.dart';
+
+/// Suggestor that updates a [setState] call to [setStateWithUpdater] in the
+/// case that the first argument is a function.
+class SetStateUpdater extends GeneralizingAstVisitor
+    with AstVisitingSuggestorMixin
+    implements Suggestor {
+  SetStateUpdater();
```
greglittlefield-wf: #nit This constructor isn't necessary; it gets declared implicitly when there are no other constructors

lib/src/component2_suggestors/setstate_updater.dart
```diff
@@ -0,0 +1,44 @@
...
+
+  @override
+  visitMethodInvocation(MethodInvocation node) {
+    super.visitMethodInvocation(node);
+
+    if (node.argumentList.arguments.isEmpty) return;
+
+    final firstArg = node.argumentList.arguments.first;
+
+    if (node.methodName.name == 'setState' && firstArg is MethodInvocation) {
```
greglittlefield-wf: Usages of updater functions are likely going to be either inline functions or tear-offs, not the result of invocations:
```dart
setState((prevState, props) {
  // return ...;
});

setState(_updater);
_updater(prevState, props) {
  // return ...;
}
```
But I would only really worry about the first case; you could do `firstArg is FunctionExpression` to easily test for that.

Would you mind updating the test cases and logic for that?

---

Also, here's a real-world example usage: https://github.com/Workiva/web_skin_dart/blob/master@%7B05-09-2019%7D/lib/src/ui_components/dropdown/trigger/multi_select/typeahead/typeahead_select.dart#L152-L168

joebingham-wk: Ah yep, good catch. I added two test cases - I couldn't think of any other exotic ways (other than the tear-off). Let me know if there are any other cases I should add coverage for!

lib/src/component2_suggestors/setstate_updater.dart
```diff
@@ -0,0 +1,45 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:analyzer/analyzer.dart';
+import 'package:codemod/codemod.dart';
+
+import '../constants.dart';
```
sydneyjodon-wk: #nit unused import

lib/src/component2_suggestors/setstate_updater.dart
```diff
@@ -0,0 +1,45 @@
...
+import '../util.dart';
```
sydneyjodon-wk: #nit unused import

lib/src/component2_suggestors/setstate_updater.dart
```diff
@@ -0,0 +1,45 @@
...
+    final firstArg = node.argumentList.arguments.first;
+
+    if (node.methodName.name == 'setState') {
+      if (firstArg is MethodInvocation &&
+          firstArg.methodName.name == 'newState') return;
+
+      if (firstArg is MethodInvocation || firstArg is FunctionExpression) {
+        int length = node.toString().indexOf('(');
+
+        yieldPatch(node.offset, node.offset + length, 'setStateWithUpdater');
```
sydneyjodon-wk: #nit maybe you could change this to use `methodName` so then you won't have to find `length`
```suggestion
        yieldPatch(node.methodName.offset, node.methodName.end, 'setStateWithUpdater');
```

lib/src/component2_suggestors/setstate_updater.dart
joebingham-wk: 💯 Ah, thank you. Good catch!

lib/src/component2_suggestors/setstate_updater.dart
joebingham-wk: That's a really good idea. In my original attempt when I only looking for a `MethodInvocation`, that would have been ideal. Now that a `FunctionExpression` is involved though, it's not guaranteed that `methodName` will exist.. If there's a real detriment to using `length` as it is, definitely let me know and that would make it worth refactoring. Otherwise (after experimenting with refactoring), it seems simpler to keep it as is. That was a suuuper good catch though 😄

sydneyjodon-wk: Oh that's true! I didn't think about it being a `FunctionExpression`.

joebingham-wk: Update: Sydney was 100% correct and I was looking at it incorrectly. That has been updated!

pubspec.yaml
```diff
@@ -28,8 +28,10 @@ dev_dependencies:
   dart_dev: ^2.0.1
   dart_style: ^1.2.0
   dependency_validator: ^1.2.3
+  meta: ^1.1.7
   mockito: ^4.0.0
   pedantic: ^1.4.0
+  react: ^4.6.1
```
aaronlademann-wf: I'm pretty new to this repo - so I'll admit I don't fully understand how dependencies relate to the expected dependencies of the repos at the time of the codemod execution, but since this codemod is targeting stuff that is within the `5.1.0-wip` branch of react, do we need a dependency override for that?

joebingham-wk: That's a really good question. I don't believe the version of react comes into play because the analyzer only needs to be included for [testing within the base branch](https://github.com/Workiva/over_react_codemod/blob/CPLAT-5212_component_2_migration/class_annotation/test/component2_suggestors/class_name_and_annotation_migrator_test.dart#L127) (from what I can tell). From what I can tell, the analyzer could see the import statement outside of `lib` but react wasn't a dependency, so it was failing and just needed something in the `pubspec`. If you'd rather my override it just to be sure though, I'm happy to do that @aaronlademann-wf!

test/component2_suggestors/setState_updater_test.dart
```diff
@@ -0,0 +1,117 @@
...
+    });
+
+    test('is a function as the first parameter', () {
+      testSuggestor(expectedPatchCount: 1, input: '''
+          main() {
+            setState(getUpdatedState());
+          }
+        ''', expectedOutput: '''
+          main() {
+            setStateWithUpdater(getUpdatedState());
```
greglittlefield-wf: I don't think we should codemod this case.

Method invocations aren't likely going to return another function; in this case, getUpdatedState would have to be:
```dart
Map Function(Map, Map) getUpdatedState() {
  return (prevState, props) { ... };
}
```
which I don't think anyone does.

lib/src/component2_suggestors/setstate_updater.dart
```diff
@@ -0,0 +1,41 @@
...
+  @override
+  visitMethodInvocation(MethodInvocation node) {
+    super.visitMethodInvocation(node);
+
+    if (node.argumentList.arguments.isEmpty) return;
+
+    final firstArg = node.argumentList.arguments.first;
+
+    if (node.methodName.name == 'setState') {
+      if (firstArg is MethodInvocation &&
```
greglittlefield-wf: See comment on test case below; we shouldn't need to check for `MethodInvocation`s here or in the next statement.

test/component2_suggestors/setState_updater_test.dart
joebingham-wk: Ah, I see what you mean now. Thank you!

lib/src/component2_suggestors/setstate_updater.dart
joebingham-wk: 💯

PR #50, opened 2019-09-12
CPLAT-7372: Fix errors caused by trying to parse versions from dependency overrides

kealjones-wk:
## Motivation
  
An unexpected issue was found when testing the codemod. If you run the react16_upgrade when it would try and parse the version of a dependency found in `dependency_overrides`, (aka the version value was like "git:" and/or "path: ../".

## Changes
  
- Added a try that wraps the version parsing and in the catch return if we find "git:" or "path:" in the message of the error.
- Added regression tests for `dependency_overrides` with git and path overrides 

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist

### DIFF
--- a/lib/src/dart2_suggestors/pubspec_over_react_upgrader.dart
+++ b/lib/src/dart2_suggestors/pubspec_over_react_upgrader.dart
@@ -71,25 +71,32 @@ class PubspecOverReactUpgrader implements Suggestor {
     if (overReactMatch != null) {
       // over_react is already in pubspec.yaml
       final constraintValue = overReactMatch.group(2);
-      final constraint = VersionConstraint.parse(constraintValue);
+      try {
+        final constraint = VersionConstraint.parse(constraintValue);
 
-      if (shouldUpdateVersionRange(
-          targetConstraint: targetConstraint,
-          constraint: constraint,
-          shouldIgnoreMin: shouldIgnoreMin)) {
-        // Wrap the new constraint in quotes if required.
-        var newValue =
-            generateNewVersionRange(constraint, targetConstraint).toString();
+        if (shouldUpdateVersionRange(
+            targetConstraint: targetConstraint,
+            constraint: constraint,
+            shouldIgnoreMin: shouldIgnoreMin)) {
+          // Wrap the new constraint in quotes if required.
+          var newValue =
+              generateNewVersionRange(constraint, targetConstraint).toString();
 
-        if (mightNeedYamlEscaping(newValue)) {
-          newValue = '"$newValue"';
+          if (mightNeedYamlEscaping(newValue)) {
+            newValue = '"$newValue"';
+          }
+
+          // Update the version constraint to ensure a safe minimum bound.
+          yield Patch(
+              sourceFile,
+              sourceFile.span(overReactMatch.start, overReactMatch.end),
+              '  over_react: $newValue');
         }
-
-        // Update the version constraint to ensure a safe minimum bound.
-        yield Patch(
-            sourceFile,
-            sourceFile.span(overReactMatch.start, overReactMatch.end),
-            '  over_react: $newValue');
+      } catch (e) {
+        // We can skip these they are versions we dont want to mess with in this codemod
+        if (e.toString().contains('git:') || e.toString().contains('path:'))
+          return;
+        rethrow;
       }
     } else if (shouldAddDependencies) {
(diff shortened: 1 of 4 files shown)


lib/src/dart2_suggestors/pubspec_over_react_upgrader.dart
```diff
@@ -71,25 +71,32 @@ class PubspecOverReactUpgrader implements Suggestor {
...
-        // Update the version constraint to ensure a safe minimum bound.
-        yield Patch(
-            sourceFile,
-            sourceFile.span(overReactMatch.start, overReactMatch.end),
-            '  over_react: $newValue');
+      } catch (e) {
+        // We can skip these they are versions we dont want to mess with in this codemod
+        if (e.toString().contains('git:') || e.toString().contains('path:'))
+          return;
+        rethrow;
```
aaronlademann-wf: You sure you want to rethrow? Won't that result in the codemod choking out when it finds one of these?

lib/src/dart2_suggestors/pubspec_over_react_upgrader.dart
```diff
@@ -71,25 +71,32 @@ class PubspecOverReactUpgrader implements Suggestor {
...
+              '  over_react: $newValue');
         }
-
-        // Update the version constraint to ensure a safe minimum bound.
-        yield Patch(
-            sourceFile,
-            sourceFile.span(overReactMatch.start, overReactMatch.end),
-            '  over_react: $newValue');
+      } catch (e) {
+        // We can skip these they are versions we dont want to mess with in this codemod
```
joebingham-wk: #nit Could we add the why behind this? If not, maybe just clean it up a little:
```suggestion
        // We can skip these. They are versions we don't want to mess with in this codemod.
```

lib/src/react16_suggestors/pubspec_react_upgrader.dart
```diff
@@ -44,23 +44,30 @@ class PubspecReactUpdater implements Suggestor {
...
-            sourceFile.span(reactMatch.start, reactMatch.end),
-            '  react: $newValue');
+          // Update the version constraint to ensure a safe minimum bound.
+          yield Patch(
+              sourceFile,
+              sourceFile.span(reactMatch.start, reactMatch.end),
+              '  react: $newValue');
+        }
+      } catch (e) {
+        // We can skip these they are versions we dont want to mess with in this codemod
```
joebingham-wk: Same #nit as above

test/dart2_suggestors/pubspec_over_react_upgrader_test.dart
```diff
@@ -93,6 +130,43 @@ main() {
...
+              expectedPatchCount: 0,
+              shouldDartfmtOutput: false,
+              validateContents: validatePubspecYaml,
+              input: ''
+                  'dependency_overrides:\n'
+                  '  over_react:\n'
+                  '    git:\n'
+                  '      url: git@github.com:cleandart/react-dart.git\n'
+                  '      ref: 5.0.0-wip\n',
+              expectedOutput: ''
```
joebingham-wk: #nit You could leave the expected output off if you wanted as it defaults to expecting the input, but it's also more explicit stating the expected output, so just a thought either way 🤷‍♂ (if you wanted to remove it then naturally it could be for all four test cases)

lib/src/dart2_suggestors/pubspec_over_react_upgrader.dart
kealjones-wk: So if `git:` or `path:` show up in the error string, we expect those errors and don't want to do anything so we return on them; Otherwise, its a non expected error and so i rethrow, since we don't know if its expected or not.

joebingham-wk: That does bring up the possibility of adding the return statement to line 97. Effective Dart [points out](https://dart.dev/guides/language/effective-dart/style#do-use-curly-braces-for-all-flow-control-structures) that if you don't use curly braces that the `if` should be on a single line, which I do think would make it more readable to the effect that you're going for! (Sorry I didn't catch that on my pass!)

kealjones-wk: blame ddev format

kealjones-wk:
@Workiva/release-management-p

PR #57, opened 2019-09-24
CPLAT-6933 React 16 Post Rollout Codemod

joebingham-wk:
## Motivation
  
We need a codemod that can remove React 16 progress comments and update the pubspec dependencies.
## Changes
  
- Add a suggestor for removing comments
- Add a new executable
- Add tests
## Review

  

Please review: 
@aaronlademann-wf @kealjones-wk @greglittlefield-wf @sydneyjodon-wk 
### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist

### DIFF
--- a/bin/react16_post_rollout_cleanup.dart
+++ b/bin/react16_post_rollout_cleanup.dart
@@ -1,0 +1,15 @@
+// Copyright 2019 Workiva Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+export 'package:over_react_codemod/src/executables/react16_post_rollout_cleanup.dart';
--- a/lib/src/component2_suggestors/component2_utilities.dart
+++ b/lib/src/component2_suggestors/component2_utilities.dart
@@ -141,7 +141,6 @@ bool fullyUpgradableToComponent2(ClassDeclaration classNode) {
 /// * Generic parameters on component class
 /// * `@AbstractProps` in the same file
 bool canBeExtendedFrom(ClassDeclaration classNode) {
-  var a = classNode.typeParameters;
   if (classNode != null &&
       (classNode.abstractKeyword != null ||
           classNode.typeParameters != null ||
--- a/lib/src/executables/react16_post_rollout_cleanup.dart
+++ b/lib/src/executables/react16_post_rollout_cleanup.dart
@@ -36,7 +36,7 @@ void main(List<String> args) {
   final overReactVersionConstraint = VersionConstraint.parse('^3.1.0');
   final logger = Logger('over_react_codemod.fixmes');
 
-  // Strings that correlate to the React 16 comments beginning and end. Based
+  // Strings that correlate to the React 16 comments' beginning and end. Based
   // on the pattern used for the React 16 update, the comments should always
   // have these strings.
   const startingString = 'Check this box';
--- a/lib/src/executables/react16_post_rollout_cleanup.dart
+++ b/lib/src/executables/react16_post_rollout_cleanup.dart
@@ -14,12 +14,12 @@
 
 import 'dart:io';
 
-import 'package:logging/logging.dart';
 import 'package:codemod/codemod.dart';
 import 'package:over_react_codemod/src/react16_suggestors/constants.dart';
 import 'package:over_react_codemod/src/dart2_suggestors/pubspec_over_react_upgrader.dart';
 import 'package:over_react_codemod/src/react16_suggestors/comment_remover.dart';
 import 'package:over_react_codemod/src/react16_suggestors/pubspec_react_upgrader.dart';
+import 'package:over_react_codemod/src/react16_suggestors/react16_utilities.dart';
 import 'package:path/path.dart' as p;
 import 'package:pub_semver/pub_semver.dart';
 
@@ -34,12 +34,19 @@ Then, review the the changes, address any FIXMEs, and commit.
 """;
 
 void main(List<String> args) {
-  const commentStringToRemove = willBeRemovedCommentSuffix;
-
-  // Update Pubspec
   final reactVersionConstraint = VersionConstraint.parse('^5.1.0');
-  final overReactVersionConstraint =
-  VersionConstraint.parse('^3.1.0');
+  final overReactVersionConstraint = VersionConstraint.parse('^3.1.0');
+
+  final query = FileQuery.dir(
+    pathFilter: isDartFile,
+    recursive: true,
+  );
+
+  if (hasUnaddressedReact16Comment(query)) {
+    throw Exception('There are still unaddressed comments from the '
+        'React 16 upgrade codemod. These should be addressed before cleaup is'
+        ' attempted.');
+  }
 
   final pubspecYamlQuery = FileQuery.dir(
     pathFilter: (path) => p.basename(path) == 'pubspec.yaml',
@@ -58,32 +65,15 @@ void main(List<String> args) {
     changesRequiredOutput: _changesRequiredOutput,
   );
 
-  if (exitCode != 0) {
-    return;
-  }
+  if (exitCode != 0) return;
 
-  // Update Componentry
-  final query = FileQuery.dir(
-    pathFilter: isDartFile,
-    recursive: true,
-  );
   exitCode = runInteractiveCodemodSequence(
     query,
     [
-      CommentRemover(commentStringToRemove)
+      CommentRemover(react16CommentsToRemove),
     ],
     args: args,
     defaultYes: true,
     changesRequiredOutput: _changesRequiredOutput,
   );
-
-  final logger = Logger('over_react_codemod.fixmes');
-  for (var dartFile in query.generateFilePaths()) {
-    final dartSource = File(dartFile).readAsStringSync();
-    if (dartSource.contains('[ ] $manualValidationCommentSubstring')) {
-      logger.severe(
-          'over_react_codemod validation comments are unaddressed in $dartFile');
-      exitCode = 1;
-    }
-  }
 }
--- a/lib/src/executables/react16_post_rollout_cleanup.dart
+++ b/lib/src/executables/react16_post_rollout_cleanup.dart
@@ -15,6 +15,7 @@
 import 'dart:io';
 
 import 'package:codemod/codemod.dart';
+import 'package:logging/logging.dart';
 import 'package:over_react_codemod/src/react16_suggestors/constants.dart';
 import 'package:over_react_codemod/src/dart2_suggestors/pubspec_over_react_upgrader.dart';
 import 'package:over_react_codemod/src/react16_suggestors/comment_remover.dart';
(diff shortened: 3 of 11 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/executables/react16_post_rollout_cleanup.dart
```diff
@@ -0,0 +1,79 @@
...
+  final reactVersionConstraint = VersionConstraint.parse('^5.1.0');
+  final overReactVersionConstraint = VersionConstraint.parse('^3.1.0');
+
+  final query = FileQuery.dir(
+    pathFilter: isDartFile,
+    recursive: true,
+  );
+
+  if (hasUnaddressedReact16Comment(query)) {
+    throw Exception('There are still unaddressed comments from the '
```
joebingham-wk: Is `Exception` the correct error to throw here? Per the Dart documentation, it seems to fit the purpose of an exception but the documentation also discourages using `Exception`s entirely.

greglittlefield-wf: I would recommend logging a `severe` error, setting the exit code, and returning instead. Throwing an uncaught exception can be a bit ugly to the user and make it seem like something is broken with the executable.

greglittlefield-wf: *bump*

lib/src/executables/react16_post_rollout_cleanup.dart
```diff
@@ -0,0 +1,82 @@
...
+  const endingString = 'complete';
+
+  final query = FileQuery.dir(
+    pathFilter: isDartFile,
+    recursive: true,
+  );
+
+  if (hasUnaddressedReact16Comment(query, logger: logger)) {
+    throw Exception('There are still unaddressed comments from the '
+        'React 16 upgrade codemod. These should be addressed before cleaup is'
```
greglittlefield-wf:
```suggestion
        'React 16 upgrade codemod. These should be addressed before cleanup is'
```

lib/src/react16_suggestors/comment_remover.dart
```diff
@@ -0,0 +1,64 @@
...
+    super.visitCompilationUnit(node);
+
+    int startingOffset;
+    int endingOffset;
+
+    for (var comment in allComments(node.root.beginToken)) {
+      if (comment != null) {
+        final commentText = sourceFile.getText(comment.offset, comment.end);
+
+        if (commentText.contains(RegExp(startString)) &&
```
greglittlefield-wf: Do these need to be wrapped in `RegExp`?

You could also make `startString` and `endString` be of type `Pattern`, which is what `contains` accepts and both String and RegExp implement.

```diff
-  final String startString;
-  final String endString;
+  final Pattern startString;
+  final Pattern endString;
```
```suggestion
        if (commentText.contains(startString) &&
```

lib/src/react16_suggestors/comment_remover.dart
```diff
@@ -0,0 +1,64 @@
...
+            startingOffset == null) {
```
greglittlefield-wf: #supernit Generally, cheaper operations (like null checks) should go first in the `&&`, allowing the short-circuiting to skip the more expensive ones (like `.contains`).

I'm not concerned about the perf here, though, so this can stay as-is.

lib/src/react16_suggestors/comment_remover.dart
```diff
@@ -0,0 +1,64 @@
...
+
+  @override
+  visitCompilationUnit(CompilationUnit node) {
+    super.visitCompilationUnit(node);
+
+    int startingOffset;
+    int endingOffset;
+
+    for (var comment in allComments(node.root.beginToken)) {
+      if (comment != null) {
```
greglittlefield-wf: When would this be null?

lib/src/react16_suggestors/comment_remover.dart
```diff
@@ -0,0 +1,64 @@
...
+        }
+      }
+
+      if (startingOffset != null && endingOffset != null) {
+        try {
+          yieldPatch(startingOffset, endingOffset, '');
+          startingOffset = null;
+          endingOffset = null;
+        } catch (e, st) {
+          throw StateError('$e\n$st');
```
greglittlefield-wf: Why is a StateError being thrown here? As a result of overlapping patches? This could use a comment, and potentially a more helpful error message.

lib/src/react16_suggestors/constants.dart
```diff
@@ -19,9 +19,16 @@ const manualValidationCommentSubstring = 'Check this box upon manual';
...
     ' This will be removed once the transition to React 16 is complete.';
 
-const styleMapExample = '''
-    // CSS number strings are no longer auto-converted to px. Ensure values are of type `num`, or have units.
-    // Incorrect value for 'width': '40'. Correct values: 40, '40px', '4em'.''';
+const styleMapExplanation = '// CSS number strings are no longer '
+    'auto-converted to px. Ensure values are of type `num`, or have units.';
+
+const styleMapExample = '// Incorrect value for \'width\': \'40\'. Correct'
+    'values: 40, \'40px\', \'4em\'.';
```
greglittlefield-wf: This is pretty hard to read; I'm a fan of double quotes when there are a lot of single quotes. Also, you don't always have to split up strings; it's okay if they occasionally go past the line length:

```dart
const styleMapExample = 
    "// Incorrect value for 'width': '40'. Correct values: 40, '40px', '4em'.";
```


Another option is to use the multiline string behavior where leading newlines are automatically trimmed:
```dart
const styleMapExample = '''
// Incorrect value for 'width': '40'. Correct values: 40, '40px', '4em'.'''
```

lib/src/react16_suggestors/react16_utilities.dart
```diff
@@ -108,3 +112,23 @@ Iterable<Token> allComments(Token beginToken) sync* {
...
+
+/// Returns whether or not there is a React 16 upgrade comment within a
+/// project that is unaddressed.
+bool hasUnaddressedReact16Comment(FileQuery query, {Logger logger}) {
+  bool hasUnaddressedComment = false;
+
+  for (var dartFile in query.generateFilePaths()) {
+    final dartSource = File(dartFile).readAsStringSync();
+    if (dartSource.contains('[ ] $manualValidationCommentSubstring')) {
+      if (logger != null) {
```
greglittlefield-wf: #nit could be null-aware: `logger?.severe(...)`

lib/src/executables/react16_post_rollout_cleanup.dart
joebingham-wk: Nice, that does make a lot of sense. Thank you!

lib/src/react16_suggestors/comment_remover.dart
joebingham-wk: Good catch, thank you for noting it!

lib/src/react16_suggestors/comment_remover.dart
joebingham-wk: That's much nicer, I didn't know about `Pattern`s and they do make it cleaner!

lib/src/react16_suggestors/constants.dart
joebingham-wk: I had a default Webstorm setting on that autowrapped at 80. Usually it wasn't a big deal and didn't get in my way, but it's caused a handful of fun formatting moments lately.. so I turned that off haha.

But nice, good to know they can run passed the line length every now and then!

lib/src/react16_suggestors/comment_remover.dart
joebingham-wk: Ah, yeah, that's not needed. In my early TDD there were errors being thrown so I was trying various things and this was a small check I threw in there just to be sure, and then I never took it out. But you're right, `comment` should never be `null`!

lib/src/react16_suggestors/comment_remover.dart
joebingham-wk: This is actually similar to [the comment above](https://github.com/Workiva/over_react_codemod/pull/57#discussion_r330644902) - there were errors getting thrown so I included the try/catch to debug easier.

 I left it deliberately thinking it'd be nice to have if it ever did hit a case that confused it, but I didn't actually have a specific case in mind that I was weary of. That's why I chose a `StateError` - it seemed the most applicable to a general unknown failure. Does that cause more confusion than good?

lib/src/component2_suggestors/component2_utilities.dart
```diff
@@ -141,7 +141,6 @@ bool fullyUpgradableToComponent2(ClassDeclaration classNode) {
 /// * Generic parameters on component class
 /// * `@AbstractProps` in the same file
 bool canBeExtendedFrom(ClassDeclaration classNode) {
-  var a = classNode.typeParameters;
```
joebingham-wk: This was just an unused variable that was popping up in my lints. It doesn't relate to the rest of the PR at all.

lib/src/executables/react16_post_rollout_cleanup.dart
```diff
@@ -48,9 +48,11 @@ void main(List<String> args) {
   );
 
   if (hasUnaddressedReact16Comment(query, logger: logger)) {
-    throw Exception('There are still unaddressed comments from the '
-        'React 16 upgrade codemod. These should be addressed before cleaup is'
-        ' attempted.');
+    logger.severe(
+        'There are still unaddressed comments from the React 16 upgrade codemod. '
+        'These should be addressed before cleanup is attempted.');
+    exitCode = 64;
```
greglittlefield-wf: Why 64?

lib/src/react16_suggestors/comment_remover.dart
greglittlefield-wf: I think wrapping it in `StateError` doesn't help much; if there's an exception, we should just let it bubble up instead of wrapping it in a generic error.

lib/src/executables/react16_post_rollout_cleanup.dart
joebingham-wk: It's the `usage` error code. My thought was that they're using the codemod incorrectly if they still have outstanding comments to address, so out of the specific error codes it fits the best 🤷‍♂

lib/src/react16_suggestors/comment_remover.dart
joebingham-wk: 🆒 🆒 🆒 I will remove that then!

lib/src/executables/react16_post_rollout_cleanup.dart
greglittlefield-wf: I'm not sure I agree there. From https://www.freebsd.org/cgi/man.cgi?query=sysexits&apropos=0&sektion=0&manpath=FreeBSD+4.3-RELEASE&format=html
> EX_USAGE (64)
> The command was used incorrectly, e.g., with the
> wrong number of arguments, a bad flag, a bad syntax
> in a parameter, or whatever.

This case will happen during normal running of the command (e.g., running codemod for the first time, running again before going in and addressing all comments), and doesn't seem like incorrect usage.

Can we switch to `1` to match the others?

test/react16_suggestors/comment_remover_test.dart
```diff
@@ -0,0 +1,266 @@
...
+
+const manualCheckedStyleMapString = '// [x] Check this box upon manual '
+    'validation that this style map uses a valid value for the keys that are '
+    'numbers.';
+
+const manualCheckedRefString = '// [x] Check this box upon manual validation of'
+    ' this ref and its typing. $willBeRemovedCommentSuffix';
+
+main() {
+  group('DependencyOverrideUpdater', () {
```
greglittlefield-wf: This name needs to be updated

lib/src/executables/react16_post_rollout_cleanup.dart
joebingham-wk: Yep, definitely! 😄

test/react16_suggestors/comment_remover_test.dart
```diff
@@ -0,0 +1,266 @@
...
+        expectedPatchCount: 0,
+        input: '''
+          class test extends UiComponent {
+            // Hello!
+          }
+        ''',
+      );
+    });
+
+    group('when removing style map comments', () {
```
greglittlefield-wf: These tests look great; could we just add two more test cases that it doesn't remove adjacent comments (above and below) the comments it's removing?

bin/react16_post_rollout_cleanup.dart
```diff
@@ -0,0 +1,15 @@
+// Copyright 2019 Workiva Inc.
```
greglittlefield-wf: I believe this executable needs to be added to pubspec.yaml.

test/react16_suggestors/comment_remover_test.dart
joebingham-wk: 🤦‍♂

bin/react16_post_rollout_cleanup.dart
joebingham-wk: Yep, 💯. Forgot about that. Thank you!

greglittlefield-wf:
+10

PR #74, opened 2020-01-30
CPLAT-9308 Codemod Utility to tell if Something is Public API

sydneyjodon-wk:
## Motivation
  
To support CPLAT-9205, we need a way to tell whether something is a public API.
- Don't worry about generating the semver report in this ticket. We can either run `semver_audit` or download the appropriate report in the repo_commander script. And then the codemod can assume that it exists at a "report.json" file or something

- Use a local semver_audit report to tell whether a component class is part of the public API. What the report looks like:
    ```json
    {
        "exports": {
           "lib/web_skin_dart.dart/ButtonProps": {
              "type": "class",
              "grammar": {
                  "name": "ButtonProps",
                  "meta": ["@Props()"]
              }
           }
        }
    }
    ```
- Add a function (likely encapsulated in a class, or whatever makes sense):
   ```dart
   List<String> getPublicExportLocations(CompilationUnitDeclaration node) { ... }
   ```
   - Return a List instead of a bool so that so it can output helpful messages saying why a component wasn't codemodded, and also in case the codemod wants to perform conditional logic based on the name of exports.
- Make it possible to call this function repeatedly re-parsing the semver report (which can be upwards of 10MB for some repos)
   - E.g.,
      ```dart
      var helper = SemverHelper.fromReport(reportJson);
      ...
      var suggestors = [
        Suggestor1(semverHelper: helper),
        Suggestor2(semverHelper: helper),
      ];
      ```
## Changes
  
- Write `SemverHelper` class with `getPublicExportLocations` utility.
- Write tests.
- Create simple `report.json` mock to help with testing codemod suggestors in other tickets.
- Add tests to existing migrators.
## Review

  

Please review: @aaronlademann-wf @joebingham-wk @greglittlefield-wf 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -0,0 +1,45 @@
...
+    final className = node.name.name;
+    String classKey;
+
+    _exportList.forEach((key, value) {
+      if (value['type'] == 'class' && value['grammar']['name'] == className) {
+        classKey = key;
+      }
+    });
+
+    return [classKey, _exportList[classKey]];
```
greglittlefield-wf: Is returning the `classKey` necessary? If not, this API would be easier to use if it was just `_exportList[classKey]` that was returned (with specific typing).

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -0,0 +1,45 @@
...
+    var fileContents;
+    await File(path).readAsString().then((String contents) {
+      fileContents = contents;
+    });
+
+    Map decoded = jsonDecode(fileContents);
+    _exportList = decoded['exports'];
+  }
+
+  List<dynamic> getPublicExportLocations(ClassDeclaration node) {
```
greglittlefield-wf: This should have a doc comment

test/boilerplate_suggestors/props_and_state_classes_migrator_test.dart
```diff
@@ -0,0 +1,47 @@
...
+import '../util.dart';
+
+main() {
+  group('PropsAndStateClassesMigrator', () {
+    SuggestorTester testSuggestor;
+
+    setUpAll(() async {
+      final helper = SemverHelper();
+      await helper.fromReport(
+          '/Users/sydneyjodon/Documents/GitHub/over_react_codemod/lib/src/boilerplate_suggestors/test.json');
```
greglittlefield-wf: We need a version of the report tracked in Git somehow, as opposed to this hard-coded API.

To make testing easier, perhaps having a constructor that takes in the map directly would be helpful. That way, you can embed the JSON in the test, and also choose how to load the report in actual usage.

```dart
SemverHelper.fromReport(this.exportList);
```

```dart
// in the test for SemverHelper
final helper = SemverHelper({'exports': ...});
```

```dart
// in usage:
final helper = SemverHelper(jsonDecode(await File('report.json').readAsString()));
```

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -0,0 +1,45 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:analyzer/dart/ast/ast.dart';
+
+class SemverHelper {
+  static Map _exportList;
```
greglittlefield-wf: This should be an instance member so that different instances of the class don't interfere with each other.

test/boilerplate_suggestors/props_and_state_classes_migrator_test.dart
greglittlefield-wf: Also, a test for a Suggestor would be a case where we'd want to use the mock for SemverHelper as opposed to a real instance, to make test setup easier (by not having to set up the whole exports map).

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
sydneyjodon-wk: I added `classKey` to the return just because I wasn't sure what kind of information would be needed in the helpful messages and it seemed like something like `lib/web_skin_dart.dart/ButtonProps` might be useful. But I guess they would already know what file they're in so it's probably not necessary.

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -119,3 +116,29 @@ void migrateClassToMixin(ClassDeclaration node, YieldPatch yieldPatch,
...
+  }
+
+  /// Returns semver report information for [node] if it is publicly exported.
+  ///
+  /// If [node] is not publicly exported, returns `null`.
+  Map<String, dynamic> getPublicExportLocations(ClassDeclaration node) {
+    final className = node?.name?.name;
+    String classKey;
+
+    if (className == null || _exportList == null) return null;
```
joebingham-wk: It seems like we may want this to be a louder failure. Have we seen the class name be `null` before? I'd be curious to see what that code looks like haha

If these can be `null` though, I'd expect it to throw an `Exception` since we need the list and class name to really verify if the class is exported, and returning `null` could be a false negative

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -119,3 +116,29 @@ void migrateClassToMixin(ClassDeclaration node, YieldPatch yieldPatch,
...
+
+    _exportList.forEach((key, value) {
+      if (value['type'] == 'class' && value['grammar']['name'] == className) {
+        classKey = key;
+      }
+    });
```
joebingham-wk: Some of these semver reports are a little large, so I think this could be optimized a little to break once it finds a match:

```dart
for (final key in _exportList.keys) {
  final value = _exportList[key];
      
  if (value['type'] == 'class' && value['grammar']['name'] == className) {
    return value;
  }
}
```

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
greglittlefield-wf: That makes sense! If we were to do this, using a class to represent this would be preferred so that it's clear what the data being returned represents, and also so that we'd have better typing:
```dart
class ExportLocations {
  final String classKey;
  final List<String> exportList;

  ExportLocations(this.classKey, this.exportList);
}
```

But yeah, for this case we probably just need the list.

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
```diff
@@ -28,13 +28,19 @@ import 'boilerplate_utilities.dart';
...
+  final SemverHelper helper;
+
+  AdvancedPropsAndStateClassMigrator(this.helper);
+
   @override
   visitClassDeclaration(ClassDeclaration node) {
     super.visitClassDeclaration(node);
 
-    if (!shouldMigrateAdvancedPropsAndStateClass(node)) return;
+    if (!shouldMigrateAdvancedPropsAndStateClass(node, helper)) return;
```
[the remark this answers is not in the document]
greglittlefield-wf: > The main thing I'm trying to prevent is passing around the same helper instance through all the suggestors and utilities, but if that seems like a good goal then I'm open to whatever implementation works best!

Passing around the instance is actually what I'd prefer, as opposed to a global variable.

Advantages:
- Eliminates the risk of different areas of the code unintentionally interfering with each other's data (including in tests)
-Make it easier to reason about where this is used, and force us to be intentional about where we pass it

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -20,11 +20,40 @@ import 'package:over_react_codemod/src/util.dart';
...
+bool isPublic(ClassDeclaration node) {
+  assert(semverHelper != null);
+  return semverHelper.getPublicExportLocations(node) != null;
+}
+
+class SemverHelper {
+  Map _exportList;
+
+  SemverHelper(Map jsonReport) {
+    _exportList = jsonReport['exports'];
```
greglittlefield-wf: #nit According to the Dart style guide, [initializing formals are preferred](https://dart.dev/guides/language/effective-dart/usage#do-use-initializing-formals-when-possible) over initializing fields in the constructor body, and [final fields are also preferred when possible](https://dart.dev/guides/language/effective-dart/design#prefer-making-fields-and-top-level-variables-final).

So, this would become:
```dart
final Map _exportList;

SemverHelper(Map jsonReport) : _exportList = jsonReport['exports'];
```

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -20,11 +20,40 @@ import 'package:over_react_codemod/src/util.dart';
...
+  }
+
+  /// Returns semver report information for [node] if it is publicly exported.
+  ///
+  /// If [node] is not publicly exported, returns `null`.
+  Map<String, dynamic> getPublicExportLocations(ClassDeclaration node) {
+    assert(_exportList != null);
```
greglittlefield-wf: I would expect this to be done in the constructor, so that the error occurs early and reliably. FYI, you can also do that in the initializer list!
```dart
SemverHelper(Map jsonReport) : 
    _exportList = jsonReport['exports'],
    assert(_exportList != null);

File: lib/src/boilerplate_suggestors/boilerplate_utilities.dart [hunk 10]
```diff
@@ -20,11 +20,40 @@ import 'package:over_react_codemod/src/util.dart';
...
+  Map _exportList;
+
+  SemverHelper(Map jsonReport) {
+    _exportList = jsonReport['exports'];
+  }
+
+  /// Returns semver report information for [node] if it is publicly exported.
+  ///
+  /// If [node] is not publicly exported, returns `null`.
+  Map<String, dynamic> getPublicExportLocations(ClassDeclaration node) {
```
greglittlefield-wf: The name `getPublicExportLocations` implies it returns several locations, whereas this function only returns only a single result. I think that having multiple values would be helpful, so that we can tell the user all of the places a class is exported as opposed to just the first one we come accross.

test/boilerplate_suggestors/simple_props_and_state_class_migrator_test.dart
```diff
@@ -24,6 +27,12 @@ main() {
     final testSuggestor =
         getSuggestorTester(SimplePropsAndStateClassMigrator(converter));
 
+    setUpAll(() async {
+      semverHelper = SemverHelper(jsonDecode(
+          await File('test/boilerplate_suggestors/report.json')
```
greglittlefield-wf: #nit as opposed to reading from a file, we could also make this a dart constant.

```dart
const reportJson = r'''{
  "exports": {
    ...
  }
}''';
```

```dart
semverHelper = SemverHelper(jsonDecode(reportJson));
```

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
greglittlefield-wf: Also, the original AC had a signature of List<String>, where the `String` is the `key` of the `_exportLIst` map. Let's stick with that, since that's all the information we should need.

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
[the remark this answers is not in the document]
joebingham-wk: I could definitely see that, and normally I wouldn't advocate for a global variable, but my main thought is that the function which answers the question of whether a node is public or not is at the utility level, and it doesn't seem like the suggestor should need to be concerned with how it answers that question (outside of passing in the node).

Also, because this helper should receive the report for the relevant library it's running on, it seems like it makes sense for it to operate independently since any instantiation for an executable should be passing it the same report. For tests that does break down, but then the testing environment could have its own instance when needed. If you feel strongly, Greg, I think that generally global variables are bad, this is just a weird case to me 😄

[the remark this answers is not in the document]
greglittlefield-wf: I feel pretty strongly here that moving away from a top-level variable would improve the modularity of this code, as well as the predictability of how data is passed around and used.

> it seems like it makes sense for it to operate independently since any instantiation for an executable should be passing it the same report

That's true, but may not always be true. For example, what if we wanted this utility to behave slightly differently for different suggestors? Or if we wanted to run multiple sets of suggestors on different directories concurrently? Having global variables would make it much more difficult to implement these changes. To be fair, this is the weakest argument of the ones I've presented, but it is something to keep in mind.

---

I'm happy to make a PR with those changes, especially since some work was already done to move it more toward top-level variables.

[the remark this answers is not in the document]
joebingham-wk: Moving away from a global variable sounds good to me. Thanks for explaining your logic, Greg! And sorry for making ya change it in the first place Sydney 😅

PR #76, opened 2020-02-03
CPLAT-9407 Add Advanced Props Migrator

joebingham-wk:
## Motivation
  
There are advanced cases of props / state that needs to be handled differently than those which simply extend the corresponding UI class with no mixins.

## Changes
  
- Build out the `AdvancedPropsAndStateClassMigrator`
- Add tests
## Review

  

Please review: 
@aaronlademann-wf @greglittlefield-wf @sydneyjodon-wk 
### QA Checklist
    - [ ] Steps from PR author: 
        
      - [ ] CI Passes

## Merge Checklist


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
```diff
@@ -28,11 +29,51 @@ import 'boilerplate_utilities.dart';
...
+      newClassDeclarationString +=
+          '$parentClass, ${className}Mixin${hasMixins ? ',' : ';'}';
+    }
+
+    if (hasMixins) {
+      if (!extendsFromCustomClass) {
+        newClassDeclarationString += '${className}Mixin,';
+      }
+
+      newClassDeclarationString = node.withClause?.childEntities
```
greglittlefield-wf: This shouldn't have null-aware operators since hasMixins is true, meaning it will never be null. Also, if it were null, `newClassDeclarationString` would get assigned to `null`.

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
```diff
@@ -28,11 +29,51 @@ import 'boilerplate_utilities.dart';
...
+          ?.whereType<TypeName>()
```
greglittlefield-wf: Instead of `.withClause.childEntities.whereType<TypeName>()` you could just do `.withClause.mixinTypes`

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
```diff
@@ -28,11 +29,51 @@ import 'boilerplate_utilities.dart';
...
+    final extendsFromCustomClass = !extendsFromUiPropsOrUiState(node);
+    final hasMixins = node.withClause != null;
+
+    // Don't operate if the props class uses mixins and extends a custom class,
+    // unless the flag has been set.
+    if (hasMixins &&
+        extendsFromCustomClass &&
+        !shouldMigrateCustomClassAndMixins) return;
+
+    final className = stripPrivateGeneratedPrefix(node.name.toSource());
```
greglittlefield-wf: Avoid using `toSource()` when there's an alternative available that's more precise/predictable
```suggestion
    final className = stripPrivateGeneratedPrefix(node.name.name);
```

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
```diff
@@ -28,11 +29,51 @@ import 'boilerplate_utilities.dart';
...
+    var newClassDeclarationString =
+        '\n\nclass $className = Ui${className.contains('Props') ? 'Props' : 'State'} with ';
+
+    if (extendsFromCustomClass) {
+      final parentClass = node.extendsClause.superclass.toSource() + 'Mixin';
```
greglittlefield-wf: Avoid `toSource`
```suggestion
      final parentClass = node.extendsClause.superclass.name.name + 'Mixin';
```

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
```diff
@@ -28,11 +29,51 @@ import 'boilerplate_utilities.dart';
...
+    }
+
+    if (hasMixins) {
+      if (!extendsFromCustomClass) {
+        newClassDeclarationString += '${className}Mixin,';
+      }
+
+      newClassDeclarationString = node.withClause?.childEntities
+          ?.whereType<TypeName>()
+          ?.joinWithToSource(
```
greglittlefield-wf: Avoid toSource. Alternatively, you could grab the name and use standard iterable utils to construct this string

```dart
final mixinNames = node.withClause.childEntities.withClause
    .mixinTypes
    .map((type) => type.name.name);
newClassDeclarationString += '${mixinNames.join(', ')};';
```


Also, using a string buffer could potentially make the code that continually appends to this string easier to read (but that's more personal preference):
```dart
newDeclarationBuffer
  ..write(mixinNames.join(', '))
  ..write(';');

```

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -116,6 +120,36 @@ void migrateClassToMixin(ClassDeclaration node, YieldPatch yieldPatch,
     newMixinName = '${newMixinName}Mixin';
   }
 
+  if (shouldSwapParentClass) {
+    final isAPropsClass = node.name.toSource().contains('Props');
```
greglittlefield-wf: Avoid toSource
```suggestion
    final isAPropsClass = node.name.name.contains('Props');
```

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -0,0 +1,455 @@
...
+      test('it\'s an empty file', () {
+        testSuggestor(expectedPatchCount: 0, input: '');
+
+        expect(propsAndStateClassNamesConvertedToNewBoilerplate, isEmpty);
+      });
+
+      test('the class is simple', () {
+        testSuggestor(
+          expectedPatchCount: 0,
+          input: '''
```
greglittlefield-wf: #fyi (what you have here is fine) you can make this a raw string so you don't have to escape all of the $ signs:
```dart
        input: r'''
        @Factory()
        UiFactory<FooProps> Foo =
            // ignore: undefined_identifier
            $Foo;
        ...
        '''
```

joebingham-wk: Ah yep, I switched that out, it looks a lot cleaner :)

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
joebingham-wk: I added the feedback! I really like putting the logic into an extension method, so I just built the feedback into that as well, but let me know if it it's really unnecessary and I'll just move it into the suggestor 😆

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -0,0 +1,455 @@
...
+          render() {
+            return Dom.ul()(
+              Dom.li()('Foo: ', props.foo),
+              Dom.li()('Bar: ', props.bar),
+            );
+          }
+        }
+      ''',
+            expectedOutput: r'''
+      @Factory()
```
aaronlademann-wf: #supernit indentation

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -0,0 +1,455 @@
...
+            );
+          }
+        }
+      ''',
+          );
+
+          expect(propsAndStateClassNamesConvertedToNewBoilerplate, {
+            'FooProps': 'FooPropsMixin',
+            'FooState': 'FooStateMixin',
+          });
```
aaronlademann-wf: Should this be...

```diff
 expect(propsAndStateClassNamesConvertedToNewBoilerplate, {
   'FooProps': 'FooPropsMixin',
   'FooState': 'FooStateMixin',
+  'ADifferentStateClass': 'ADifferentStateClassMixin',
 });
```

Or is that piece of the map added in a portion of logic that is not exercised in this test?

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -0,0 +1,455 @@
...
+        testSuggestor(
+          expectedPatchCount: 0,
+          input: r'''
+      @Factory()
+      UiFactory<FooProps> Foo =
+          // ignore: undefined_identifier
+          $Foo;
+
+      @Props()
+      class _$FooProps extends ADifferentPropsClass with AMixin, AnotherMixin {
```
aaronlademann-wf: What about the case where `ADifferentPropsClass` is the one with mixins?

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -0,0 +1,455 @@
...
+        testSuggestorWithFlag(
+          expectedPatchCount: 14,
+          input: r'''
+        @Factory()
+        UiFactory<FooProps> Foo =
+            // ignore: undefined_identifier
+            $Foo;
+
+        @Props()
+        class _$FooProps extends ADifferentPropsClass with AMixin, AnotherMixin {
```
aaronlademann-wf: What about the cases where:

* `ADifferentPropsClass` is the one that has mixins
* The mixins either found in this `with` clause, or on the `with` clause of `ADifferentPropsClass` implement more than just `UiProps`? 
    * _Those interfaces would need to get mixed into the final concrete props class as well._

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -116,6 +119,33 @@ void migrateClassToMixin(ClassDeclaration node, YieldPatch yieldPatch,
     newMixinName = '${newMixinName}Mixin';
   }
 
+  if (shouldSwapParentClass) {
```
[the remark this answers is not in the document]
joebingham-wk: I would have expected so too, but git managed to resolve it! I did have to move the couple of if statements into the concrete implementation section, though.

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
aaronlademann-wf: Offline resolution - always add a `// FIXME` comment when a class extends from a class that is not `UiProps`.

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
aaronlademann-wf: Offline resolution - always add a `// FIXME` comment when a class extends from a class that is not `UiProps`.

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
joebingham-wk: Yep, it was excluded because that class didn't get converted in this test case, but we're assuming it would be

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -0,0 +1,471 @@
...
+        testSuggestor(
+          expectedPatchCount: 0,
+          input: r'''
+          @Factory()
+          UiFactory<FooProps> Foo =
+              // ignore: undefined_identifier
+              $Foo;
+    
+          @Props()
+          class _$FooProps extends ADifferentPropsClass with AMixin, AnotherMixin {
```
aaronlademann-wf: I think this case should get migrated, with a `// FIXME` comment just like the case where there is a custom superclass with no mixins.

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
```diff
@@ -42,26 +41,30 @@ class AdvancedPropsAndStateClassMigrator extends GeneralizingAstVisitor
...
     final className = stripPrivateGeneratedPrefix(node.name.name);
     final newDeclarationBuffer = StringBuffer()
+      ..write('\n\n')
+      // Write a fix me comment if this class extends a custom class
+      ..write(!extendsFromCustomClass
+          ? ''
+          : '''
+          // FIXME: Verify that:
+          //   1. $convertedParentClass does not implement additional mixins.
+          //   2. any other mixins found here do not implement additional mixins.
```
greglittlefield-wf: These instructions are pretty vague on needs to be done, and so is the term "implement".

I would  expect the first bullet to read as
> Ensure that all mixins used by ADifferentPropsClass are also mixed into this class.
ADifferentPropsClassMixin shouldn't implement other mixins, right? All of those should have shifted to ADifferentPropsClass?

> any other mixins found here do not implement additional mixins
What case is this talking about? Wouldn't the analyzer warn you about this? If so, we should just say "Fix any analyzer warnings on this class about missing mixins"

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -138,8 +138,8 @@ void migrateClassToMixin(ClassDeclaration node, YieldPatch yieldPatch,
 extension IterableAstUtils on Iterable<NamedType> {
   /// Utility to join an `Iterable` based on the `name` of the `name` field
   /// rather than the `toString()` of the object.
-  String joinByName({String startingString, String endingString}) {
-    final itemString = map((t) => t.name.name).join(', ');
+  String joinByName({String startingString, String endingString, String seperator}) {
```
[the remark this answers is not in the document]
joebingham-wk: 🤦‍♂

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -137,19 +136,15 @@ void migrateClassToMixin(ClassDeclaration node, YieldPatch yieldPatch,
...
-    forEach((e) {
-      final isTheLastItem = last.name.toSource() == e.name.toSource();
-
-      string += '${e.name.toSource()}${isTheLastItem ? '' : ', '}';
-    });
-
-    string += endingString ?? '';
-
-    return string;
+  String joinByName({String startingString, String endingString}) {
```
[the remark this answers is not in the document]
joebingham-wk: That is correct, I can remove them here! Nothing uses them

PR #78, opened 2020-02-04
CPLAT-9412 Add annotation remover

aaronlademann-wf:
## Motivation
We need a migrator that removes unnecessary `@Factory()`, `@Props()`, `@State()` and `@Component2()` annotations.

## Changes
1. Add logic for the migrator
2. Write tests

## Review

  

Please review: @joebingham-wk @sydneyjodon-wk @greglittlefield-wf 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist

### DIFF
--- a/lib/src/boilerplate_suggestors/annotations_remover.dart
+++ b/lib/src/boilerplate_suggestors/annotations_remover.dart
@@ -16,12 +16,78 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:codemod/codemod.dart';
 
-/// Suggestor that looks for @Props, @State, and @Component2 and removes them.
+import 'boilerplate_utilities.dart';
+
+/// Suggestor that looks for `@Props()`, `@State()`, `@Component2()` and `@Factory()` annotations, and removes them
+/// as long as they do not contain any arguments.
+///
+/// > NOTE: `@PropsMixin()` and `@StateMixin()` annotations are removed via [PropsMixinMigrator].
 class AnnotationsRemover extends GeneralizingAstVisitor
     with AstVisitingSuggestorMixin
     implements Suggestor {
   @override
   visitAnnotatedNode(AnnotatedNode node) {
     super.visitAnnotatedNode(node);
+    if (node.metadata.isEmpty) return;
+
+    final annotationToRemove =
+        node.metadata.singleWhere(_annotationIsRelevant, orElse: () => null);
+
+    // --- Short Circuit Conditions --- //
+    if (annotationToRemove == null) return;
+    if (annotationToRemove.arguments.arguments.isNotEmpty) return;
+    if (!_propsClassWasConvertedToNewBoilerplate(node)) return;
+
+    // --- Migrate --- //
+    yieldPatch(annotationToRemove.offset,
+        node.firstTokenAfterCommentAndMetadata.offset, '');
+  }
+
+  static const _relevantAnnotationNames = [
+    'Factory',
+    'Props',
+    'State',
+    'Component2',
+  ];
+
+  bool _annotationIsRelevant(Annotation annotation) =>
+      _relevantAnnotationNames.contains(annotation.name.name);
+
+  bool _nodeHasRelevantAnnotation(AnnotatedNode node) =>
+      node.metadata.any(_annotationIsRelevant);
+
+  bool _nodeHasAnnotationWithName(AnnotatedNode node, String annotationName) =>
+      node.metadata.any((annotation) => annotation.name.name == annotationName);
+
+  String _getNameOfPropsClassThatMayHaveBeenConverted(AnnotatedNode node) {
+    if (_nodeHasAnnotationWithName(node, 'Factory')) {
+      TopLevelVariableDeclaration factoryNode = node;
+      return getPropsClassNameFromFactoryDeclaration(factoryNode);
+    } else if (_nodeHasAnnotationWithName(node, 'State')) {
+      ClassOrMixinDeclaration stateNode = node;
+      return stateNode.name.name.replaceFirst('State', 'Props');
+    } else if (_nodeHasAnnotationWithName(node, 'Component2')) {
+      ClassDeclaration componentNode = node;
+      return componentNode.name.name.replaceFirst('Component', 'Props');
+    }
+
+    return null;
+  }
+
+  bool _propsClassWasConvertedToNewBoilerplate(AnnotatedNode node) {
+    if (_nodeHasAnnotationWithName(node, 'Props')) {
+      // Its the props class
+      return node is MixinDeclaration &&
+          propsAndStateClassNamesConvertedToNewBoilerplate
+              .containsKey(node.name.name);
+    } else if (_nodeHasRelevantAnnotation(node)) {
+      // Not a props class, but is a UiComponent-related class with an annotation.
+      final analogousPropsMixinOrClassName =
+          _getNameOfPropsClassThatMayHaveBeenConverted(node);
+      return propsAndStateClassNamesConvertedToNewBoilerplate
+          .containsKey(analogousPropsMixinOrClassName);
+    }
+
+    return false;
   }
 }
--- a/lib/src/boilerplate_suggestors/annotations_remover.dart
+++ b/lib/src/boilerplate_suggestors/annotations_remover.dart
@@ -60,15 +60,12 @@ class AnnotationsRemover extends GeneralizingAstVisitor
       node.metadata.any((annotation) => annotation.name.name == annotationName);
 
   String _getNameOfPropsClassThatMayHaveBeenConverted(AnnotatedNode node) {
-    if (_nodeHasAnnotationWithName(node, 'Factory')) {
-      TopLevelVariableDeclaration factoryNode = node;
-      return getPropsClassNameFromFactoryDeclaration(factoryNode);
-    } else if (_nodeHasAnnotationWithName(node, 'State')) {
-      ClassOrMixinDeclaration stateNode = node;
-      return stateNode.name.name.replaceFirst('State', 'Props');
-    } else if (_nodeHasAnnotationWithName(node, 'Component2')) {
-      ClassDeclaration componentNode = node;
-      return componentNode.name.name.replaceFirst('Component', 'Props');
+    if (node is TopLevelVariableDeclaration) {
+      return getPropsClassNameFromFactoryDeclaration(node);
+    } else if (node is ClassDeclaration) {
+      return node.name.name.replaceFirst('Component', 'Props');
+    } else if (node is ClassOrMixinDeclaration) {
+      return node.name.name.replaceFirst('State', 'Props');
     }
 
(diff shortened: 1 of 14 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/boilerplate_suggestors/annotations_remover.dart
```diff
@@ -16,12 +16,78 @@ import 'package:analyzer/dart/ast/ast.dart';
...
+    if (_nodeHasAnnotationWithName(node, 'Factory')) {
+      TopLevelVariableDeclaration factoryNode = node;
+      return getPropsClassNameFromFactoryDeclaration(factoryNode);
+    } else if (_nodeHasAnnotationWithName(node, 'State')) {
+      ClassOrMixinDeclaration stateNode = node;
+      return stateNode.name.name.replaceFirst('State', 'Props');
+    } else if (_nodeHasAnnotationWithName(node, 'Component2')) {
+      ClassDeclaration componentNode = node;
+      return componentNode.name.name.replaceFirst('Component', 'Props');
+    }
```
joebingham-wk: Instead of casting the node after doing a check, was there a reason that just doing a type check didn't work? Something like:

```
String _getNameOfPropsClassThatMayHaveBeenConverted(AnnotatedNode node) {
 if (node is TopLevelVariableDeclaration) {
    return getPropsClassNameFromFactoryDeclaration(node);
  } else if (node is ClassDeclaration) {
    return node.name.name.replaceFirst('Component', 'Props');
  } else if (node is ClassOrMixinDeclaration) {
    return node.name.name.replaceFirst('State', 'Props');
  }

  return null;
}
```

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -119,3 +120,14 @@ void migrateClassToMixin(ClassDeclaration node, YieldPatch yieldPatch,
...
+
+/// Returns the name of the props class for a given [factoryDeclaration].
+String getPropsClassNameFromFactoryDeclaration(
+    TopLevelVariableDeclaration factoryDeclaration) {
+  return factoryDeclaration.variables.type.childEntities
+      .whereType<TypeArgumentList>()
+      .single
+      .arguments
+      .single
+      .toString();
```
joebingham-wk: I believe `toSource()` is preferred over `.toString()`
```suggestion
      .toSource();
```

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:meta/meta.dart';
```
joebingham-wk: #nit unused import

lib/src/boilerplate_suggestors/annotations_remover.dart
```diff
@@ -16,12 +16,75 @@ import 'package:analyzer/dart/ast/ast.dart';
...
 class AnnotationsRemover extends GeneralizingAstVisitor
     with AstVisitingSuggestorMixin
     implements Suggestor {
   @override
   visitAnnotatedNode(AnnotatedNode node) {
     super.visitAnnotatedNode(node);
+    if (node.metadata.isEmpty) return;
+
+    final annotationToRemove =
+        node.metadata.singleWhere(_annotationIsRelevant, orElse: () => null);
```
greglittlefield-wf: #nit We should use `firstWhere` instead of `singleWhere` where possible since it doesn't throw upon finding duplicate matches (and it also short-circuits, which is a small perf bonus)

lib/src/boilerplate_suggestors/annotations_remover.dart
```diff
@@ -16,12 +16,75 @@ import 'package:analyzer/dart/ast/ast.dart';
...
+/// as long as they do not contain any arguments.
+///
+/// > NOTE: `@PropsMixin()` and `@StateMixin()` annotations are removed via [PropsMixinMigrator].
 class AnnotationsRemover extends GeneralizingAstVisitor
     with AstVisitingSuggestorMixin
     implements Suggestor {
   @override
   visitAnnotatedNode(AnnotatedNode node) {
     super.visitAnnotatedNode(node);
+    if (node.metadata.isEmpty) return;
```
greglittlefield-wf: #nit this check isn't necessary since it's handled by the `singleWhere` case (`orElse` is called for empty iterables)

lib/src/boilerplate_suggestors/annotations_remover.dart
```diff
@@ -16,12 +16,75 @@ import 'package:analyzer/dart/ast/ast.dart';
...
+        node.metadata.singleWhere(_annotationIsRelevant, orElse: () => null);
+
+    // --- Short Circuit Conditions --- //
+    if (annotationToRemove == null) return;
+    if (annotationToRemove.arguments.arguments.isNotEmpty) return;
+    if (!_propsClassWasConvertedToNewBoilerplate(node)) return;
+
+    // --- Migrate --- //
+    yieldPatch(annotationToRemove.offset,
+        node.firstTokenAfterCommentAndMetadata.offset, '');
```
greglittlefield-wf: This could potentially remove other code (e.g., other annotations) between the relevant annotation and the class declarations. Dartfmt will handle adjusting newlines/whitespace as needed.
```suggestion
        annotationToRemove.end, '');
```

lib/src/boilerplate_suggestors/annotations_remover.dart
```diff
@@ -16,12 +16,75 @@ import 'package:analyzer/dart/ast/ast.dart';
...
+
+/// Suggestor that looks for `@Props()`, `@State()`, `@Component2()` and `@Factory()` annotations, and removes them
+/// as long as they do not contain any arguments.
+///
+/// > NOTE: `@PropsMixin()` and `@StateMixin()` annotations are removed via [PropsMixinMigrator].
 class AnnotationsRemover extends GeneralizingAstVisitor
     with AstVisitingSuggestorMixin
     implements Suggestor {
   @override
   visitAnnotatedNode(AnnotatedNode node) {
```
greglittlefield-wf: #nit We don't need to traverse over all annotated nodes, which includes things like local variables and function arguments. `visitCompilationUnitMember` would be more specific.

lib/src/boilerplate_suggestors/annotations_remover.dart
```diff
@@ -16,12 +16,75 @@ import 'package:analyzer/dart/ast/ast.dart';
...
+  bool _annotationIsRelevant(Annotation annotation) =>
+      _relevantAnnotationNames.contains(annotation.name.name);
+
+  bool _nodeHasRelevantAnnotation(AnnotatedNode node) =>
+      node.metadata.any(_annotationIsRelevant);
+
+  bool _nodeHasAnnotationWithName(AnnotatedNode node, String annotationName) =>
+      node.metadata.any((annotation) => annotation.name.name == annotationName);
+
+  String _getNameOfPropsClassThatMayHaveBeenConverted(AnnotatedNode node) {
```
greglittlefield-wf: :nice:

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -119,3 +120,14 @@ void migrateClassToMixin(ClassDeclaration node, YieldPatch yieldPatch,
...
+
+/// Returns the name of the props class for a given [factoryDeclaration].
+String getPropsClassNameFromFactoryDeclaration(
+    TopLevelVariableDeclaration factoryDeclaration) {
+  return factoryDeclaration.variables.type.childEntities
+      .whereType<TypeArgumentList>()
+      .single
+      .arguments
+      .single
+      .toSource();
```
greglittlefield-wf: #nit This could break if the type argument isn't specified. Also, avoid single
```dart
  return factoryDeclaration.variables.type.childEntities
      .whereType<TypeArgumentList>()
      .firstOrNull
      ?.arguments
      ?.first
      ?.toSource();
```

```dart
extension<E> on Iterable<E> {
  E firstOrNull => isEmpty ? null : first;
}

File: lib/src/boilerplate_suggestors/annotations_remover.dart [hunk 10]
```diff
@@ -16,12 +16,75 @@ import 'package:analyzer/dart/ast/ast.dart';
...
+    // --- Migrate --- //
+    yieldPatch(annotationToRemove.offset,
+        node.firstTokenAfterCommentAndMetadata.offset, '');
+  }
+
+  static const _relevantAnnotationNames = [
+    'Factory',
+    'Props',
+    'State',
+    'Component2',
```
greglittlefield-wf: Also relevant: `AbstractProps`, `AbstractState`.

lib/src/boilerplate_suggestors/annotations_remover.dart
```diff
@@ -16,12 +16,75 @@ import 'package:analyzer/dart/ast/ast.dart';
...
+  bool _nodeHasAnnotationWithName(AnnotatedNode node, String annotationName) =>
+      node.metadata.any((annotation) => annotation.name.name == annotationName);
+
+  String _getNameOfPropsClassThatMayHaveBeenConverted(AnnotatedNode node) {
+    if (node is TopLevelVariableDeclaration) {
+      return getPropsClassNameFromFactoryDeclaration(node);
+    } else if (node is ClassDeclaration) {
+      return node.name.name.replaceFirst('Component', 'Props');
+    } else if (node is ClassOrMixinDeclaration) {
+      return node.name.name.replaceFirst('State', 'Props');
```
greglittlefield-wf: These should be more precise replacements instead of just a String pattern. For example, with teh current code, `GraphFormStateWrapperProps` would become `GraphFormPropsWrapperProps`.
```suggestion
      return node.name.name.replaceFirst(RegExp(r'State$'), 'Props');
```

test/boilerplate_suggestors/annotations_remover_test.dart
```diff
@@ -0,0 +1,229 @@
...
+import 'package:test/test.dart';
+
+import '../util.dart';
+
+main() {
+  group('AnnotationsRemover', () {
+    final testSuggestor = getSuggestorTester(AnnotationsRemover());
+
+    tearDown(() {
+      propsAndStateClassNamesConvertedToNewBoilerplate = {};
```
greglittlefield-wf: I know this wasn't added in this PR, so it shouldn't be addressed here, but at some point we probably want to change it from a global variable and either pass it around to the relevant objects/functions or encapsulate it with the relevant utility methods. FYI @joebingham-wk 

In doing so, we'll 
- eliminate the risk of different areas of the code unintentionally interfering with each other's data (including in tests like this) 
- make it easier to reason about where the data is used
- decouple code that reads/writes to this map from the consuming code

aaronlademann-wf: @greglittlefield-wf PR up to address this: https://github.com/Workiva/over_react_codemod/pull/79

PR #79, opened 2020-02-06
Use utility class to contain map of converted classNames

aaronlademann-wf:
## Motivation
We want to change the global variable being used to keep track of props / state classes that have been converted to the new boilerplate by sibling migrators to a variable within an object that can be passed around by utilities / migrators that need it.

In doing so, we'll

* eliminate the risk of different areas of the code unintentionally interfering with each other's data (including in tests like this)
* make it easier to reason about where the data is used
* decouple code that reads/writes to this map from the consuming code

## Changes
* Create a `ClassToMixinConverter` class, which contains a `convertedClassNames` getter to access the private `_convertedClassNames` field.
* Create a single instance of this class within `main()` of `boilerplate_upgrade.dart`, and pass the instance around to the migrators that make use of it as constructor arguments.

FYA @sydneyjodon-wk @joebingham-wk @greglittlefield-wf

### DIFF
--- a/lib/src/executables/boilerplate_upgrade.dart
+++ b/lib/src/executables/boilerplate_upgrade.dart
@@ -16,6 +16,7 @@ import 'dart:io';
 
 import 'package:codemod/codemod.dart';
 import 'package:over_react_codemod/src/boilerplate_suggestors/annotations_remover.dart';
+import 'package:over_react_codemod/src/boilerplate_suggestors/boilerplate_utilities.dart';
 import 'package:over_react_codemod/src/boilerplate_suggestors/props_meta_migrator.dart';
 import 'package:over_react_codemod/src/boilerplate_suggestors/simple_props_and_state_class_migrator.dart';
 import 'package:over_react_codemod/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart';
@@ -38,6 +39,8 @@ void main(List<String> args) {
     },
     recursive: true,
   );
+
+  final classToMixinConverter = ClassToMixinConverter();
 
   // General plan:
   //  - Things that need to be accomplished (very simplified)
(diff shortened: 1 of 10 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/executables/boilerplate_upgrade.dart
```diff
@@ -79,11 +82,11 @@ void main(List<String> args) {
...
-      SimplePropsAndStateClassMigrator(),
-      AdvancedPropsAndStateClassMigrator(),
-      PropsMixinMigrator(),
-      PropsMetaMigrator(),
-      AnnotationsRemover(),
+      SimplePropsAndStateClassMigrator(classToMixinConverter),
+      AdvancedPropsAndStateClassMigrator(classToMixinConverter),
+      PropsMixinMigrator(classToMixinConverter),
+      PropsMetaMigrator(classToMixinConverter),
+      AnnotationsRemover(classToMixinConverter),
```
joebingham-wk: This class still needs the constructor to be updated

test/boilerplate_suggestors/props_meta_migrator_test.dart
```diff
@@ -20,25 +20,26 @@ import '../util.dart';
 
 main() {
   group('PropsMetaMigrator', () {
-    final testSuggestor = getSuggestorTester(PropsMetaMigrator());
+    final converter = ClassToMixinConverter();
+    final testSuggestor = getSuggestorTester(PropsMetaMigrator(converter));
 
     tearDown(() {
-      propsAndStateClassNamesConvertedToNewBoilerplate = {};
+      converter.setConvertedClassNames({});
```
greglittlefield-wf: #nit Instead of sharing this converter and having to worry about cleaning up after it, it'd be nice to have a fresh instance each time.

PR #80, opened 2020-02-11
CPLAT-9576 Consume SemverHelper

sydneyjodon-wk:
Change base back to `CPLAT-9205-new-boilerplate-codemod` once #82 is merged in.
## Motivation
  
CPLAT-9308 added a utility to tell if something is public API, and we have started to consume it in several places. However, the consumption is not complete, and we need to finish it up.

- If a user provides a command-line flag that allows overriding semver, `--treat-all-components-as-private`:
      - instantiate a special instance of `SemverHelper`, e.g., something like `SemverHelper.alwaysPrivate();` that always returns empty results for `getPublicExportLocations`

- Attempt to load semver report from semver_report.json and use it to instantiate a SemverHelper.
    - Log whether the file existed. If it did not, log a message instructing the user to perform a semver report before rerunning the codemod, or to use a command line flag (e.g., `--treat-all-components-as-private`) to skip checking against semver:
        ```
        pub global activate semver_audit --hosted-url=https://pub.workiva.org
        pub global run semver_audit generate > semver_report.json
        ```
    - If the report file did not exist:
        - instantiate a special instance of `SemverHelper`, e.g., something like `SemverHelper.alwaysPublic();` that always returns non-empty results for `getPublicExportLocations` (This could be a dummy value, like `"semver report not available; assuming this to be public"`)

- Ensure codemods that should not operate on components/classes that are publicly exported:
    - Are consuming SemverHelper correctly.
    - Are emitting warning messages saying a component wasn't codemodded "because it was exported from: $locations"
## Changes
  
- Add `--treat-all-components-as-private` flag
- Update tests to include flag and invalid file
- Add utility functions to comment public export locations and update existing comments
- Add tests for comments
## Review

  

Please review: @greglittlefield-wf @joebingham-wk @aaronlademann-wf 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -413,3 +468,88 @@ extension IterableAstUtils on Iterable<NamedType> {
...
+      existingExportLocationsCommentToken.end,
+      '',
+    );
+  }
+}
+
+/// Returns AST token for [commentToGet] from the comments above [node].
+///
+/// If [commentToGet] does not exist above [node], returns null.
+Token getComment(AstNode node, SourceFile sourceFile, String commentToGet,
```
greglittlefield-wf: I think this should live in the same file as hasComment (and hasComment could probably be DRY'd up to use the same code as `getComment`, since they're almost identical).

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
```diff
@@ -30,14 +30,24 @@ class AdvancedPropsAndStateClassMigrator extends GeneralizingAstVisitor
...
-    if (!shouldMigrateAdvancedPropsAndStateClass(node)) return;
+    if (shouldAddPublicExportLocationsAdvancedClassComment(
+        node, semverHelper)) {
+      addPublicExportLocationsComment(
+          node, sourceFile, semverHelper, yieldPatch);
+    }
+
+    if (!shouldMigrateAdvancedPropsAndStateClass(node, semverHelper)) return;
+
+    removePublicExportLocationsComment(
```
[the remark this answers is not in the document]
aaronlademann-wf: Good question... 

I added quite a bit of logic in https://github.com/Workiva/over_react_codemod/pull/82 that associates "FIX ME" comments with migration decisions.  IMO - the stuff in this PR involving the adding/removing of comments could be removed for now - and then we could circle back to make the comment added using the logic in #82 have more informative content _(like where the member is exported from, etc)_

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -12,30 +12,71 @@
...
+/// assumes all classes passed to [getPublicExportLocations] are private
+/// (see: [SemverHelper.alwaysPrivate] constructor).
+Future<SemverHelper> getSemverHelper(String path,
+    {bool shouldTreatAllComponentsAsPrivate = false}) async {
+  if (shouldTreatAllComponentsAsPrivate) {
+    return SemverHelper.alwaysPrivate();
+  } else {
+    final file = File(path);
+
+    if (await file.exists()) {
```
aaronlademann-wf: FYI... you don't have to mess with futures if you use the "sync" version... 

```suggestion
    if (file.existsSync()) {
```

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -12,30 +12,71 @@
...
+      return SemverHelper(jsonDecode(await file.readAsString()));
```
aaronlademann-wf: Same thing here...

```suggestion
      return SemverHelper(jsonDecode(await file.readAsStringSync()));
```

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -23,29 +22,62 @@ import 'boilerplate_utilities_test.dart';
...
+  });
+}
 
-    group('does not operate when', () {
-      test('it\'s an empty file', () {
-        testSuggestor(expectedPatchCount: 0, input: '');
+void AdvancedPropsAndStateClassMigratorTestHelper({
+  String path,
+  bool shouldTreatAllComponentsAsPrivate,
+  bool isValidFilePath,
```
joebingham-wk: #nit these could have default parameters, making the invocation a little simpler

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -23,29 +22,62 @@ import 'boilerplate_utilities_test.dart';
...
+}) {
+  final converter = ClassToMixinConverter();
+  final semverHelper = getSemverHelper(path,
+      shouldTreatAllComponentsAsPrivate: shouldTreatAllComponentsAsPrivate);
+  SuggestorTester testSuggestor;
+
+  setUpAll(() async {
+    testSuggestor = getSuggestorTester(
+        AdvancedPropsAndStateClassMigrator(converter, await semverHelper));
```
joebingham-wk: Was there a reason the `await` for `semverHelper` is on this line rather than line 57? I usually see `await`s on the line of the instantiation, rather than the usage.

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -1492,16 +2187,155 @@ void main() {
...
+                      'ADifferentPropsClass': 'ADifferentPropsClassMixin',
+                      'FooPropsMixin': 'FooPropsMixin',
+                    });
         });
       });
     });
   });
 }
+
+exportLocationsCommentTestHelper({
```
joebingham-wk: This just needs a return type
```suggestion
void exportLocationsCommentTestHelper({
```

test/boilerplate_suggestors/boilerplate_utilities_test.dart
```diff
@@ -12,59 +12,14 @@
...
-    "lib/another_file.dart/ButtonProps": {
-      "type": "class",
-      "grammar": {
-        "name": "ButtonProps",
-        "meta": ["@Props()"]
-      }
-    }
-  }
-}''';
+String exportLocationsComment(List<String> locations) =>
```
joebingham-wk: I'm not sure this is the best home for this utility, but it also doesn't really have a clear home haha. I'd probably either it make a new file in test/boilerplate_suggestors called "utils" or add it to the general test utils, since this file is more for testing the utilities from the boilerplate codemods and not so much the home of testing utils.

test/boilerplate_suggestors/boilerplate_utilities_test.dart
```diff
@@ -317,49 +272,101 @@ void main() {
...
+        );
+      });
+    });
+  });
+}
+
+void semverUtilitiesTestHelper({
+  String path,
+  bool shouldTreatAllComponentsAsPrivate,
+  bool isValidFilePath,
```
joebingham-wk: same #supernit as from the advanced_migrator test; you could add default parameters here and I feel like the invocations are a little cleaner, but either way!

test/boilerplate_suggestors/boilerplate_utilities_test.dart
```diff
@@ -317,49 +272,101 @@ void main() {
...
+  test('if props class is not in export list', () {
+    final input = '''
+      @Props()
+      class _\$FooProps extends UiProps{
+        String foo;
+        int bar;
+      }
+    ''';
+
+    CompilationUnit unit = parseString(content: input).unit;
```
joebingham-wk: This could be final rather than a type annotation ([effective dart](https://dart.dev/guides/language/effective-dart/design#avoid-type-annotating-initialized-local-variables)). 

The same variable is also used in the next test, so it could be lifted out of the tests and cleaned up with a `tearDown()` (in that case, keeping the type annotation)

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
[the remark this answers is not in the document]
sydneyjodon-wk: I remove the comments here because at this point, the class is going to be updated so any existing comments explaining why the class wasn't updated should be removed. `removePublicExportLocationsComment` only yields a patch if the comment exists.

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
sydneyjodon-wk: Nice!

test/boilerplate_suggestors/boilerplate_utilities_test.dart
```diff
@@ -317,49 +272,101 @@ void main() {
...
-            'lib/another_file.dart/ButtonProps'
-          ]);
-          expect(isPublic(classNode), true);
-        });
-      });
+            'lib/another_file.dart/ButtonProps',
+          ]
+        : [reportNotAvailableComment];
+
+    CompilationUnit unit = parseString(content: input).unit;
```
joebingham-wk: See comment above about type annotation or lifting it out.

test/boilerplate_suggestors/boilerplate_utilities_test.dart
```diff
@@ -317,49 +272,101 @@ void main() {
...
+    expect(unit.declarations.whereType<ClassDeclaration>().length, 1);
+
+    unit.declarations.whereType<ClassDeclaration>().forEach((classNode) {
+      expect(semverHelper.getPublicExportLocations(classNode),
+          shouldTreatAllComponentsAsPrivate ? isEmpty : expectedOutput);
+      expect(isPublic(classNode, semverHelper),
+          shouldTreatAllComponentsAsPrivate ? false : true);
```
joebingham-wk: This could be simplified to:
```suggestion
          !shouldTreatAllComponentsAsPrivate);
```

or, instead of `false` / `true`, the matchers `isFalse` or `isTrue` could be used

test/boilerplate_suggestors/props_mixin_migrator_test.dart
```diff
@@ -12,22 +12,54 @@
...
+        isValidFilePath: false,
+      );
+    });
+  });
+}
+
+PropsMixinMigratorTestHelper({
+  String path,
+  bool shouldTreatAllComponentsAsPrivate,
+  bool isValidFilePath,
```
joebingham-wk: Same #supernit about default parameters

test/boilerplate_suggestors/props_mixin_migrator_test.dart
```diff
@@ -12,22 +12,54 @@
...
+      PropsMixinMigratorTestHelper(
+        path: 'test/boilerplate_suggestors/does_not_exist.json',
+        shouldTreatAllComponentsAsPrivate: false,
+        isValidFilePath: false,
+      );
+    });
+  });
+}
+
+PropsMixinMigratorTestHelper({
```
joebingham-wk: This could use a return type
```suggestion
void PropsMixinMigratorTestHelper({
```

test/boilerplate_suggestors/props_mixin_migrator_test.dart
```diff
@@ -12,22 +12,54 @@
...
-    final testSuggestor = getSuggestorTester(PropsMixinMigrator(converter));
+    final semverHelper = getSemverHelper(path,
+        shouldTreatAllComponentsAsPrivate: shouldTreatAllComponentsAsPrivate);
+    SuggestorTester testSuggestor;
 
-    setUpAll(() {
-      semverHelper = SemverHelper(jsonDecode(reportJson));
+    setUpAll(() async {
+      testSuggestor =
+          getSuggestorTester(PropsMixinMigrator(converter, await semverHelper));
```
joebingham-wk: I also noted above about the usage of `await` here versus in the instantiation - so just noting it here as well

lib/src/dart2_suggestors/props_and_state_companion_class_remover.dart
```diff
@@ -24,8 +24,8 @@ import '../util.dart';
 class PropsAndStateCompanionClassRemover extends RecursiveAstVisitor
     with AstVisitingSuggestorMixin
     implements Suggestor {
-  bool shouldRemoveCompanionClassFor(
-          ClassDeclaration candidate, CompilationUnit node) =>
+  bool shouldRemoveCompanionClassFor(ClassDeclaration candidate,
+          ClassDeclaration companion, CompilationUnit node) =>
       true;
```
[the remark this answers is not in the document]
sydneyjodon-wk: I added this because there was an existing test for `'when the stubbed "companion" class(es) are publicly exported'` and so it seemed like it should be checking `companion.isPublic()` instead of `candidate.isPublic()`. But it looks like Aaron changed it to `ClassToMixinConverter.convertedClasses` like you said which makes it a lot simpler so I'll revert my changes here.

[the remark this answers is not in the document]
sydneyjodon-wk: Actually, I can't remove the `SemverHelper` usage in this suggestor right now because `shouldMigrateSimplePropsAndStateClass` and `shouldMigrateAdvancedPropsAndStateClass` need the `SemverHelper` argument. I put notes to remove this once Aaron's changes that remove those two utility function usages are merged in.

lib/src/boilerplate_suggestors/stubbed_props_and_state_class_remover.dart
```diff
@@ -18,15 +18,21 @@ import 'package:over_react_codemod/src/boilerplate_suggestors/advanced_props_and
 import 'package:over_react_codemod/src/boilerplate_suggestors/simple_props_and_state_class_migrator.dart';
 import 'package:over_react_codemod/src/dart2_suggestors/props_and_state_companion_class_remover.dart';
 
+import 'boilerplate_utilities.dart';
+
 /// Suggestor that removes every companion class for props and state classes, as
 /// they were only temporarily required for backwards-compatibility with Dart 1.
 class StubbedPropsAndStateClassRemover
     extends PropsAndStateCompanionClassRemover implements Suggestor {
+  final SemverHelper semverHelper;
```
sydneyjodon-wk: This is necessary because the current implementation of `StubbedPropsAndStateClassRemover` uses utility functions that require `SemverHelper` to be passed in. Once PR #82 (which updates the implementation and removes those utility function usages) is merged in, this will no longer be needed and can be removed.

test/boilerplate_suggestors/stubbed_props_and_state_class_remover_test.dart
```diff
@@ -12,24 +12,19 @@
...
 main() {
   group('StubbedPropsAndStateClassRemover', () {
-    final testSuggestor = getSuggestorTester(
-      StubbedPropsAndStateClassRemover(),
-    );
-
-    setUpAll(() {
-      semverHelper = SemverHelper(jsonDecode(reportJson));
-    });
+    final semverHelper = getSemverHelper(
```
sydneyjodon-wk: This will be removed when #82 (which removes the need for `SemverHelper` in `StubbedPropsAndStateClassRemover`) is merged in.

test/boilerplate_suggestors/boilerplate_utilities_test.dart
sydneyjodon-wk: lol you're very right

test/boilerplate_suggestors/boilerplate_utilities_test.dart
sydneyjodon-wk: Yeah I didn't know where to put it either haha. I moved it to the general test utils file, but if anyone thinks it should be in its own file or somewhere else let me know.

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -406,8 +466,9 @@ class ClassToMixinConverter {
     if (node.extendsClause?.extendsKeyword != null) {
       // --- Convert concrete props/state class to a mixin --- //
 
-      yieldPatch(node.name.token.offset,
-          node.name.token.offset + privateGeneratedPrefix.length, '');
+      if (node.name.name != newMixinName) {
+        yieldPatch(node.name.token.offset, node.name.token.end, newMixinName);
+      }
```
aaronlademann-wf: This doesn't seem related to the semver stuff... is this intentional, or a bad merge possibly?

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -12,16 +12,27 @@
...
+
+    group('with --treat-all-components-as-private flag', () {
+      AdvancedPropsAndStateClassMigratorTestHelper(
+        shouldTreatAllComponentsAsPrivate: true,
+      );
+    });
+  });
+}
+
+void AdvancedPropsAndStateClassMigratorTestHelper({
```
aaronlademann-wf: #nit only classes should be named using `UpperCamelCase`

```suggestion
void advancedPropsAndStateClassMigratorTestHelper({
```

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -259,6 +277,80 @@ void main() {
...
 
+      test('the semver report does not exist, but does add a FIXME comment',
+          () {
+        testSuggestor(isValidSemverReportFilePath: false)(
+          expectedPatchCount: shouldTreatAllComponentsAsPrivate ? 0 : 2,
+          input: '''
+          $factoryDecl
+  
+          @Props()
+          class $propsClassName extends ADifferentStateClass {
```
aaronlademann-wf:
```suggestion
          class $propsClassName extends ADifferentPropsClass {
```

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -259,6 +277,80 @@ void main() {
...
+          //      pub global activate semver_audit --hosted-url=https://pub.workiva.org
+          //      pub global run semver_audit generate 2> semver_report.json
+          //   2. Remove this FIXME comment.
+          //   3. Re-run the migration script:
+          //      pub run over_react_codemod:boilerplate_upgrade
+          //
+          // Alternatively, remove this FIXME comment and re-run the  migration script 
+          // with the following flag to assume all components are not publicly exported:
+          // pub run over_react_codemod:boilerplate_upgrade --treat-all-components-as-private'''}
+          class $propsClassName extends ADifferentStateClass {
```
aaronlademann-wf:
```suggestion
          class $propsClassName extends ADifferentPropsClass {
```

test/boilerplate_suggestors/props_mixin_migrator_test.dart
```diff
@@ -12,23 +12,42 @@
...
+    group('with invalid file path', () {
+      PropsMixinMigratorTestHelper(
+        path: 'test/boilerplate_suggestors/does_not_exist.json',
+        isValidFilePath: false,
+      );
     });
+  });
+}
+
+void PropsMixinMigratorTestHelper({
```
aaronlademann-wf: Only classes should be `UpperCamelCase`

```suggestion
void propsMixinMigratorTestHelper({
```

test/boilerplate_suggestors/simple_props_and_state_class_migrator_test.dart
```diff
@@ -12,25 +12,43 @@
...
+    group('with invalid file path', () {
+      SimplePropsAndStateClassMigratorTestHelper(
+        path: 'test/boilerplate_suggestors/does_not_exist.json',
+        isValidFilePath: false,
+      );
+    });
+  });
+}
+
+void SimplePropsAndStateClassMigratorTestHelper({
```
aaronlademann-wf: Only classes should be `UpperCamelCase`

```suggestion
void simplePropsAndStateClassMigratorTestHelper({
```

PR #81, opened 2020-02-12
CPLAT-9565 Address edge cases for boilerplate codemods

aaronlademann-wf:
## Motivation
All of the codemods to support migration to the new over_react boilerplate have been merged into the integration branch, but we still needed to run them for the most complex use-cases _(e.g. WSD)_ to find out what edge cases our logic missed.

Here is what I discovered / fixed:

1. https://github.com/Workiva/over_react_codemod/pull/81/commits/4ddc3bf81093659dda0969fb118564171127f5c9 Preserve any generic type args that are present on the superclass when converting a class to a mixin.
    * Before:
      ```diff
      - class SomeProps extends SuperProps<SomeType, SomeOtherType> {}
      + mixin SomeProps on UiProps<SomeType, SomeOtherType> with SuperPropsMixin {}
      ```
    * After:
      ```diff
      - class SomeProps extends SuperProps<SomeType, SomeOtherType> {}
      + mixin SomeProps on UiProps with SuperPropsMixin<SomeType, SomeOtherType> {}
      ```
1. https://github.com/Workiva/over_react_codemod/pull/81/commits/5c0a16a15424e96ce8555b6c3f0fdf61f218fe1f Use "known" converted name of superclass instead of assuming `Mixin` has been appended to the name.
    * Before:
      ```diff
      // When SuperProps is not found within `converter.convertedClassNames.keys`
      - class SomeProps extends SuperProps {}
      + mixin SomeProps on UiProps with SuperPropsMixin {}
      ```
    * After:
      ```diff
      // When SuperProps is not found within `converter.convertedClassNames.keys`
      - class SomeProps extends SuperProps {}
      + mixin SomeProps on UiProps with SuperProps {}
      ```
1. https://github.com/Workiva/over_react_codemod/pull/81/commits/7a3c2759181daab52ac6187637dbf84efa2f1e1e, https://github.com/Workiva/over_react_codemod/pull/81/commits/dfab8dc091305b4ba3d34ee1d15761c1e8e9ddc8 and https://github.com/Workiva/over_react_codemod/pull/81/commits/a141f4c3eab84f5cab455c0d4e16dcb5e3c9fc44 Remove `const` context(s) from declarations involving `propsMeta.forMixin()` since that new constructor is not a const like `SomePropsClass.meta` was.
1. https://github.com/Workiva/over_react_codemod/pull/81/commits/7231a456fc6d7b4905afda2882f8355a2f70f859 and https://github.com/Workiva/over_react_codemod/pull/81/commits/03461115a3358619a5d5dc6ad17d20d619ad17cb Do not create a mixin if it will be an unnecessary duplicate.
    * Before:
      ```diff
      // When the class has no members
      mixin SomePropsMixin on UiProps {
        bool foo;
      }
  
      - class SomeProps extends UiProps with SomePropsMixin {}
      // woops... this already exists
      + mixin SomePropsMixin on UiProps {}
      // woops... two mixins
      + class SomeProps = UiProps with SomePropsMixin, SomePropsMixin;
  
      // ----------------------
      // When the class has members
      mixin SomePropsMixin on UiProps {
        bool foo;
      }
  
      - class SomeProps extends UiProps with SomePropsMixin {
      -  String bar;
      - }
      // woops... this already exists, and now this prop key is in no-man's land
      + mixin SomePropsMixin on UiProps {
      +  String bar;
      + }
      // woops... two mixins
      + class SomeProps = UiProps with SomePropsMixin, SomePropsMixin;
      ```
    * After:
      ```diff
      // When the class has no members
      mixin SomePropsMixin on UiProps {
        bool foo;
      }
  
      - class SomeProps extends UiProps with SomePropsMixin {}
      + class SomeProps = UiProps with SomePropsMixin;
  
      // ----------------------
      // When the class has members
      mixin SomePropsMixin on UiProps {
        bool foo;
        // This prop key has been moved into the existing mixin
        String bar;
      }
  
      class SomeProps = UiProps with SomePropsMixin;
      ```
1. Remove generated mixin names from the `with` clause when the mixin has been converted, and leave the `// ignore` comments intact for generated mixin names when they have not been converted.
    * Before:
      ```diff
      // When `ConvertedMixin` is found within `converter.convertedClassNames.keys`,
      // but `UnconvertedMixin` is not
      - class SomeProps = UiProps
          with ConvertedMixin,
      -        // ignore: mixin_of_non_class, undefined_class
              $ConvertedMixin,
              UnconvertedMixin,
      -        // ignore: mixin_of_non_class, undefined_class
              $UnconvertedMixin;
      ```
    * After:
      ```diff
      // When `ConvertedMixin` is found within `converter.convertedClassNames.keys`,
      // but `UnconvertedMixin` is not
      - class SomeProps = UiProps
          with ConvertedMixin,
      -        // ignore: mixin_of_non_class, undefined_class
      -        $ConvertedMixin,
              UnconvertedMixin,
              // ignore: mixin_of_non_class, undefined_class
              $UnconvertedMixin;
      ```
1. https://github.com/Workiva/over_react_codemod/pull/81/commits/9d8ab4c685132ed3c2c7391fd1b0712333a5d9e6, https://github.com/Workiva/over_react_codemod/pull/81/commits/995a3d225aceaaf44ecb040814db9bfe050e4218 Don't make abstract props classes concrete, and convert them to interfaces instead
    * Before:
      ```diff
      - abstract class SomeAbstractProps extends UiProps {}
      + mixin SomeAbstractPropsMixin on UiProps {}
      + class SomeAbstractProps = UiProps with SomeAbstractPropsMixin {}
      ```
    * After:
      ```diff
      - abstract class SomeAbstractProps extends UiProps {}
      + mixin SomeAbstractPropsMixin on UiProps {}
      + abstract class SomeAbstractProps implements SomeAbstractPropsMixin {}
      ```

## Changes
Update suggestor logic and tests to address the edge cases enumerated above.

## Review

  

Please review: @greglittlefield-wf @joebingham-wk @sydneyjodon-wk 

### QA Checklist
    - [ ] Steps from PR author:
        
## Merge Checklist


aviary-wf:
## Security Insights

## Action Items
---

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -211,12 +253,44 @@ class ClassToMixinConverter {
       if (shouldSwapParentClass) {
         yieldPatch(
             node.extendsClause.superclass.name.offset,
-            node.extendsClause.superclass.name.end,
+            node.extendsClause.superclass.end,
```
aaronlademann-wf: This ensures that if the old superclass was parameterized, that those parameters don't end up on `UiProps` / `UiState`.

lib/src/executables/boilerplate_upgrade.dart
```diff
@@ -81,12 +82,13 @@ void main(List<String> args) {
   exitCode = runInteractiveCodemodSequence(
     query,
     <Suggestor>[
+      PropsMixinMigrator(classToMixinConverter),
```
aaronlademann-wf: We need this to run first so that the `AdvancedPropsAndStateClassMigrator` can check for duplicate mixin names before creating one.

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
```diff
@@ -42,33 +42,36 @@ class AdvancedPropsAndStateClassMigrator extends GeneralizingAstVisitor
...
+    final classTypeArgs = node.typeParameters ?? '';
     final newDeclarationBuffer = StringBuffer()
       ..write('\n\n')
       // Write a fix me comment if this class extends a custom class
       ..write(!extendsFromCustomClass
           ? ''
           : '''
           // FIXME:
-          //   1. Ensure that all mixins used by $parentClassName are also mixed into this class.
+          //   1. Ensure that all mixins used by ${parentClassName}Mixin are also mixed into this class.
```
greglittlefield-wf: Why was this changed? ${parentClassName}Mixin can't "use" any mixins

Perhaps the wording isn't super clear, and it should be changed to:
```suggestion
          //   1. Ensure that all mixins used by the old ${parentClassName} are also mixed into this class.
```

test/boilerplate_suggestors/props_meta_migrator_test.dart
```diff
@@ -58,17 +58,30 @@ main() {
...
+            // field at the top of the class!
+            // ignore: undefined_identifier, undefined_class, const_initialized_with_non_constant_value
+            static const PropsMeta meta = _\$metaForFooPropsMixin;
+            
+            String foo;
+          }
+              
           @Component2()
           class FooComponent extends UiComponent2<FooProps> {
+            static const foo = FooProps.meta;
```
greglittlefield-wf: Curious: does this really happen often enough to justify a codemod?

lib/src/boilerplate_suggestors/props_meta_migrator.dart
```diff
@@ -38,6 +38,11 @@ class PropsMetaMigrator extends GeneralizingAstVisitor
 
   PropsMetaMigrator(this.converter);
 
+  // Utility to keep track of when a `const` keyword is removed from the left side of a `TypedLiteral`
+  // so that we can prevent "overlapping" patches when a literal contains more than one `PrefixedIdentifier`
+  // - which causes the codemod script to crash.
+  static final Map<TypedLiteral, bool> _literalsWithConstRemoved = {};
```
greglittlefield-wf: This doesn't need to be `static`, and could have its scope limited; it's only really applicable within a single traversal of a given `VariableDeclaration`.

Also, having this be an `Expando` would probably be preferable so that this Map doesn't retain any memory unnecessarily across codemods:
```suggestion
  final Expando<bool> _constRemovedFlag = Expando();
  bool _hasLiteralHadConstRemoved(TypedLiteral literal) = _constRemovedFlag[literal] ?? false;
```

lib/src/boilerplate_suggestors/props_meta_migrator.dart
```diff
@@ -48,7 +50,13 @@ class PropsMetaMigrator extends GeneralizingAstVisitor
     super.visitPrefixedIdentifier(node);
 
     if (node.identifier.name == 'meta') {
-      if (converter.convertedClassNames.containsKey(node.prefix.name)) {
+      final propsClassWithMetaWasConverted =
+          converter.convertedClassNames.containsKey(node.prefix.name);
+      final componentWithMetaUsageIsComponent2 =
```
greglittlefield-wf: This could use a small code comment to say why this check is necessary

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -632,6 +632,61 @@ void main() {
...
+            ''',
+              expectedOutput: r'''
+              @AbstractProps()
+              mixin AbstractBlockPropsMixin on UiProps implements BlockClassHelperMapView, LayoutPropsMixin, BlockPropsMixin, // ignore: mixin_of_non_class, undefined_class
+                      $BlockPropsMixin {
+                String foo;
+                int bar;
+              } 
+              
+              abstract class AbstractBlockProps = UiProps
```
greglittlefield-wf: We should really never have a props class that's abstract, since we can't extend from it with the new boilerplate. In this case, I'd expect us to eliminate the abstract class.

This case may be difficult to handle entirely in codemod, so I'd be down to just emit a FIXME for this case

greglittlefield-wf: Oh wait, is this case handled by another suggestor? If so, maybe we should test this more functionally with an AggregateSuggestor, since the intermediate states aren't as important

lib/src/executables/boilerplate_upgrade.dart
greglittlefield-wf: Can we capture this in a code comment?
```suggestion
      // We need this to run first so that the AdvancedPropsAndStateClassMigrator 
      // can check for duplicate mixin names before creating one.
      PropsMixinMigrator(classToMixinConverter),
```

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -256,7 +258,26 @@ class ClassToMixinConverter {
       }
 
       if (node.withClause != null) {
-        yieldPatch(node.withClause.offset, node.withClause.end, '');
+        // Convert the with clause to an implements clause since the mixin
+        // is how consumers will "extend" props moving forward. These mixins
+        // are moved to the new concrete class via the `AdvancedPropsAndStateClassMigrator`,
+        // but if they are not also implemented here, the new props mixin won't have all
+        // the keys that the old class / abstract class had... which will be a breaking
+        // change for consumers without a workaround/path forward.
```
greglittlefield-wf: I'm not sure I agree that we should always implement these. They shouldn't always be needed, and we shouldn't clutter the boilerplate unnecessarily for cases where it's really just the concrete props class consuming another mixin that's unrelated to the rest of the props.

The conversion from a class to mixin is already a breaking change, which is fine since in most cases we're only operating on public APIs.

Anyone using the typing outside of the component should be using the concrete props class, which will have everything mixed in, so we don't need to worry about those cases.

The only case I can think of is if a mixin is overriding/implementing a prop from an interface, which should be far less common. A less intrusive way could be to check for `@override` annotations on props, and add a FIXME saying to add missing implements clauses if necessary

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -278,10 +352,60 @@ String getPropsClassNameFromFactoryDeclaration(
...
+
+  return converter.convertedClassNames[originalClassName] ?? originalClassName;
+}
+
 extension IterableAstUtils on Iterable<NamedType> {
   /// Utility to join an `Iterable` based on the `name` of the `name` field
   /// rather than the `toString()` of the object.
-  String joinByName({String separator}) {
-    return map((t) => t.name.name).join('${separator ?? ','} ');
+  String joinByName(
```
greglittlefield-wf: This method now does much more than what it did originally, and shou,d have its name and doc comment updated.

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
```diff
@@ -211,12 +253,44 @@ class ClassToMixinConverter {
...
+            yieldPatch(node.withClause.offset, node.withClause.end,
+                ' implements $mixinNames');
+          }
+        } else if (node.implementsClause != null) {
+          // only `implements` clause exists
+          yieldPatch(node.implementsClause.end, node.implementsClause.end,
+              ', $mixinNames');
+        } else {
+          // no `with` or `implements` clause exist
+          yieldPatch(node.leftBracket.offset - 1, node.leftBracket.offset - 1,
```
greglittlefield-wf: It's not safe to assume that there's a whitespace before the left bracket. This should just be
```suggestion
          yieldPatch(node.leftBracket.offset, node.leftBracket.offset,
```

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -421,6 +421,156 @@ void main() {
...
+            mixin FooPropsMixin on UiProps {
+              String foo;
+              int bar;
+            }
+            
+            // FIXME:
+            //   1. Ensure that all mixins used by ADifferentPropsClassMixin are also mixed into this class.
+            //   2. Fix any analyzer warnings on this class about missing mixins        
+            class FooProps extends UiProps with ADifferentPropsClassMixin, FooPropsMixin {
+              String baz;
```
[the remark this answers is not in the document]
aaronlademann-wf: ok. I was wondering about that - but wasn't sure so I did it this way first since it was simpler.  I'll update the codemod to move any prop keys into the existing mixin in this case.

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
aaronlademann-wf: Yes - the codemod used to remove `abstract` altogether... I'll just revert this change.

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
aaronlademann-wf: @greglittlefield-wf If we don't implement the classes that used to be mixed in - in the new mixin - then conversions like this won't work:

__Existing structure__:

```dart
// --- abstract ---

@AbstractProps()
abstract class AbstractBlockProps extends UiProps
    with LayoutPropsMixin, BlockPropsMixin
    implements BlockClassHelperMapView {}
@AbstractComponent2()
abstract class AbstractBlockComponent<T extends AbstractBlockProps> extends UiComponent2<T> with LayoutMixin<T>, BlockMixin<T> {}

// --- concrete implementation ---

@Props()
class BlockProps extends AbstractBlockProps {}
@Component2()
class BlockComponent extends AbstractBlockComponent<BlockProps> {}
```

__After codemod__:

```dart
// --- abstract ---

mixin AbstractBlockPropsMixin on UiProps {}

class AbstractBlockProps = UiProps with AbstractBlockPropsMixin, 
    LayoutPropsMixin, BlockPropsMixin
    implements BlockClassHelperMapView;

abstract class AbstractBlockComponent<T extends AbstractBlockProps> extends UiComponent2<T> with LayoutMixin<T>, BlockMixin<T> {}

// --- concrete implementation ---

mixin BlockPropsMixin on UiProps {}

class BlockProps = UiProps with BlockPropsMixin, AbstractBlockPropsMixin {}

// Analysis error here: BlockProps does not implement AbstractBlockProps
class BlockComponent extends AbstractBlockComponent<BlockProps> {}
```

The last line above is the problem. IMO - we're really just dealing with WSD at this point - and a certain level of verbosity is expected.  And in this case - adding those interfaces to the mixins is necessary to make the components work in their current structure.  We can / should certainly phase this out over time... but for the initial conversion, it would be nice for the vast majority of things to "just work" with minimal intervention required - since there will be a few things that will take A LOT of intervention no matter how good the codemods are.

greglittlefield-wf: Yeah, I'm worried more about non-WSD usages, and would prefer to not have unnecessary `implements` clauses that clutter things up and potentially confuse consumers. 

Could we conditionally only add implements clauses when we're the component extends abstract components?

```dart
class BlockProps = UiProps 
    with BlockPropsMixin, AbstractBlockPropsMixin 
    implements AbstractBlockProps;
```

That way, it should only be added to the cases that need it.


---

Side note: in this case, we really shouldn't be making a concrete `AbstractBlockProps` class; it will never be extended from or used, so there's effectively no difference between mixing in or implementing something. How about we just make them "interfaces"
```dart
abstract class AbstractBlockProps implements 
    AbstractBlockPropsMixin, 
    LayoutPropsMixin,
    BlockPropsMixin,
    BlockClassHelperMapView;
```

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
aaronlademann-wf: nevermind - i will leave them abstract and make them interfaces as suggested here: https://github.com/Workiva/over_react_codemod/pull/81#discussion_r379032898

test/boilerplate_suggestors/props_meta_migrator_test.dart
aaronlademann-wf: 🤷‍♂ its done though... so... does it matter?

lib/src/boilerplate_suggestors/boilerplate_utilities.dart
aaronlademann-wf: Implemented this in https://github.com/Workiva/over_react_codemod/pull/81/commits/995a3d225aceaaf44ecb040814db9bfe050e4218

test/boilerplate_suggestors/props_meta_migrator_test.dart
greglittlefield-wf: Sorry, I meant that purely out of curiosity. If it did matter I'd be more direct 😜

PR #72, opened 2020-01-28
CPLAT-9205 Codemod for New Boilerplate

sydneyjodon-wk:
## Motivation
New over_react component boilerplate is coming soon, and our consumers deserve an easy way to make use of it!
   
## Changes
- [x] A migrator that acts on simple `@Props()` / `@State()` classes
  (#73)
- [x] A migrator that acts on `@Props()` / `@State()`  classes with advanced inheritance 
  (#76, #79, #81, #82)
- [x] A migrator that acts on standalone `@PropsMixin()` / `@StateMixin()` classes
  (#77)
- [x] A migrator that removes the public _(un-prefixed)_ "companion" classes
  (#75)
- [x] Utilities to determine if a class is public API before performing a migration
  (#74, #80, #83)
- [x] A migrator that removes `@Props`, `@State()`, `@PropsMixin()`, `@StateMixin()`, `@Factory()` and `@Component2()` annotations that are not required by the new boilerplate
  (#78)

#### Release Notes
Add codemod to assist over_react consumers in their migration to the new boilerplate changes.

## Review
_[See CONTRIBUTING.md][contributing-review-types] for more details on review types (+1 / QA +1 / +10) and code review process._

  

Please review: 

### QA Checklist
- [ ] Tests were updated and provide good coverage of the changeset and other affected code
- [ ] Manual testing was performed if needed
    - [ ] Steps from PR author: 
        
      - [ ] Run the boilerplate codemod on web_skin_dart and verify that the codemod updates props and state classes as expected.
        - `pub global run over_react_codemod:boilerplate_upgrade`
    - [ ] Anything falling under manual testing criteria [outlined in CONTRIBUTING.md][contributing-manual-testing]

## Merge Checklist
While we perform many automated checks before auto-merging, some manual checks are needed:
- [ ] A Client Platform member has reviewed these changes
- [ ] There are no unaddressed comments _- this check can be automated if reviewers use the "Request Changes" feature_
- [ ] _For release PRs -_ Version metadata in Rosie comment is correct

[contributing-review-types]: https://github.com/Workiva/over_react_codemod/blob/master/CONTRIBUTING.md#review-types
[contributing-manual-testing]: https://github.com/Workiva/over_react_codemod/blob/master/CONTRIBUTING.md#manual-testing-criteria

PR #82, opened 2020-02-19
CPLAT-9677 Improve class migration short-circuit logic + consumer communication

aaronlademann-wf:
> **NOTE:** I'd recommend just reading through the `advanced_props_and_state_migrator_test.dart` file without attempting to make sense of the diff.

## Motivation
Currently, when the codemod runs the `AdvancedPropsAndStateClassMigrator`, the migrator does not short-circuit if a custom superclass, or one or more mixins is not present in the map of "known converted classes" that the migrators update as classes are converted.

This is done because we have no way to control the order in which classes are visited by the suggestor.  This means that just because a superclass or mixin is not present in the map of "known converted classes" at the time the subclass is visited - does not mean it will never be converted.

We also need to improve how we communicate to the consumer when our codemod short-circuits because of public API detection / superclasses from external libs.

## Changes
1. Expand the role of the `ClassToMixinConverter` class to record visits, as well as record whether the class that was visited was converted.
    1. This allows us to short-circuit other migrations when a class that another class extends from / mixes in: 
        1. Was visited, but not converted
        1. Was never visited (meaning it lives in an external lib)
1. Run the `AdvancedPropsAndStateClassMigrator` twice during the migration.
    1. On the first run, short-circuit - but do not add FIXME comments when an unknown/unconverted superclass or mixin is encountered.
    2. On the second run, short-circuit AND add FIXME comments when an unknown/unconverted superclass or mixin is encountered.
1. Add a `--convert-classes-with-external-superclasses` flag that can be set to force the conversion of classes that extend from or mix in classes that do not live in the same library.
1. Add a `MigrationDecision` class that is responsible for determining when short-circuits occur during migrations, as well as storing a `reason` comment that we will apply to keep the consumers informed about what they can do to either migrate the classes manually, or to make some adjustments before running the migration script again.
1. Move the `StubbedPropsAndStateClassRemover` so that it runs with the knowledge of which classes have been converted.
1. Misc. logic updates to handle additional edge cases while testing the migrators in `graph_ui`.
1. Add crap-tons of tests.

## Review

  

Please review: @joebingham-wk @sydneyjodon-wk @greglittlefield-wf 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist

### DIFF
--- a/test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
+++ b/test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
@@ -1504,6 +1504,117 @@ void main() {
             });
           });
 
+          group('that is abstract based on annotation only (edge case)', () {
+            test('with members of its own', () {
+              const input = r'''
+                  @PropsMixin()
+                  abstract class BreadcrumbPathPropsMixin {}
+                  
+                  @AbstractProps()
+                  class _$AbstractBreadcrumbPathProps extends AbstractGraphFormProps
+                      with
+                          BreadcrumbPathPropsMixin,
+                          // ignore: mixin_of_non_class, undefined_class
+                          $BreadcrumbPathPropsMixin {
+                    String foo;
+                    int bar;
+                  }
+                  
+                  @AbstractComponent2()
+                  class AbstractBreadcrumbPathComponent<T extends AbstractBreadcrumbPathProps> 
+                      extends UiComponent2<T> {}
+                ''';
+
+              testSuggestor(visitedClassNames: {
+                'BreadcrumbPathPropsMixin': 'BreadcrumbPathPropsMixin',
+                'AbstractGraphFormProps': 'AbstractGraphFormProps',
+              })(
+                expectedPatchCount: 7,
+                input: input,
+                expectedOutput: r'''
+                @PropsMixin()
+                abstract class BreadcrumbPathPropsMixin {}
+              
+                @AbstractProps()
+                mixin AbstractBreadcrumbPathPropsMixin on UiProps {
+                  String foo;
+                  int bar;
+                }
+
+                // FIXME:
+                //   1. Ensure that all mixins used by AbstractGraphFormProps are also mixed into this class.
+                //   2. Fix any analyzer warnings on this class about missing mixins.
+                @AbstractProps()
+                abstract class AbstractBreadcrumbPathProps 
+                    implements
+                        AbstractGraphFormProps,
+                        AbstractBreadcrumbPathPropsMixin,
+                        BreadcrumbPathPropsMixin {}
+
+                @AbstractComponent2()
+                class AbstractBreadcrumbPathComponent<T extends AbstractBreadcrumbPathProps> 
+                    extends UiComponent2<T> {}
+              ''',
+              );
+
+              expect(converter.visitedNames, {
+                'BreadcrumbPathPropsMixin': 'BreadcrumbPathPropsMixin',
+                'AbstractGraphFormProps': 'AbstractGraphFormProps',
+                'AbstractBreadcrumbPathProps':
+                    'AbstractBreadcrumbPathPropsMixin',
+              });
+            });
+
+            test('with no members', () {
+              const input = r'''
+                  @PropsMixin()
+                  abstract class BreadcrumbPathPropsMixin {}
+                  
+                  @AbstractProps()
+                  class _$AbstractBreadcrumbPathProps extends AbstractGraphFormProps
+                      with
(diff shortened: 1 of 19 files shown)


aviary-wf:
## Security Insights

## Action Items
---

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -22,43 +22,228 @@ import '../util.dart';
...
-    final testSuggestor =
-        getSuggestorTester(AdvancedPropsAndStateClassMigrator(converter));
+    SuggestorTester testSuggestor({
+      bool convertClassesWithExternalSuperclass = false,
+      Map<String, String> visitedClassNames = const {},
+    }) {
+      runCount++;
+      final tester = getSuggestorTester(AdvancedPropsAndStateClassMigrator(
+          converter,
+          treatUnvisitedClassesAsExternal: runCount > 1,
```
sydneyjodon-wk:
```suggestion
          treatUnvisitedClassesAsExternal: runCount % 2 == 0,
```
I might not be fully understanding this, but it looks like every time the codemod is run, the first `AdvancedPropsAndStateClassMigrator` has `treatUnvisitedClassesAsExternal` set to false and the second `AdvancedPropsAndStateClassMigrator` has `treatUnvisitedClassesAsExternal` set to true.  A couple tests below are simulating running the codemod twice so it might be more realistic to alternate setting `treatUnvisitedClassesAsExternal` to true and add an additional run of `AdvancedPropsAndStateClassMigrator` to the second codemod run. I don't know if it makes much of a difference in the tests, but just a thought.

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -70,610 +255,1387 @@ void main() {
...
+          });
+
+          test(
+              'which then gets removed from the declaration that is converted to a mixin, and replaced '
+              'with updated instructions on the new concrete class declaration when the script '
+              'is re-ran with the --convert-classes-with-external-superclasses flag set',
+              () {
+            // Run it a third time - this time simulating `--convert-classes-with-external-superclasses`
+            // being set - which allows conversion of external superclasses
+            testSuggestor(convertClassesWithExternalSuperclass: true)(
```
sydneyjodon-wk: One of the examples of the second run of the codemod as mentioned in the comment above.
```suggestion
            testSuggestor()(expectedPatchCount: 0, input: expectedOutputWithExternalSuperclassReasonComment);
            testSuggestor(convertClassesWithExternalSuperclass: true)(
```

aaronlademann-wf: @sydneyjodon-wk the testSuggestor is already being run twice before this in the `setUp` of the test.

sydneyjodon-wk: Yes, that's why I'm confused as to why it isn't run twice here as well. In `boilerplate_upgrade.dart`, it looks like every time the codemod is run, AdvancedPropsAndStateClassMigrator is run twice. The first time not considering the flag and the second time with the flag. 

It seems like the test setup is running the codemod once (so running `AdvancedPropsAndStateClassMigrator` twice like you said). So I’m just wondering why in this case when simulating running the codemod again, `AdvancedPropsAndStateClassMigrator` is only run once?

sydneyjodon-wk: But I guess the tests are still passing without the added run because `converter.visitedNames` isn't set back to empty in tearDown. I just didn’t know if it would be more realistic to run the suggester twice for each run of the codemod.

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -767,238 +1725,353 @@ void main() {
...
-                  Dom.li()('Foo: ', props.foo),
-                  Dom.li()('Bar: ', props.bar),
-                );
-              }
-            }
+
+            @Props()
+            class $publicPropsClassName = UiProps
+                with ${publicPropsClassName}Mixin, ConvertedMixin, UnconvertedMixin, // ignore: mixin_of_non_class, undefined_class
+                \$UnconvertedMixin;
```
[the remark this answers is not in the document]
greglittlefield-wf: Moved to main PR: https://github.com/Workiva/over_react_codemod/pull/72/files#r386543133

greglittlefield-wf:
@rmconsole-wf

greglittlefield-wf:
@Workiva/release-management-p

PR #89, opened 2020-04-13
CPLAT-10515 Preserve consumed props behavior defaults

greglittlefield-wf:
## Motivation
Since over_react's new boilerplate will be consuming all mixed in props by defaults (via https://github.com/Workiva/over_react/pull/485), we need to update the codemod to preserve the same set of consumed props when converting legacy boilerplate.

## Changes
**1\. Update migrators to preserve this behavior via `consumedProps` impl updates:**
- simple_props_and_state_migrator - No changes necessary.

  These classes will result in only one props mixin being used, so the default behavior change still results in the single mixin being consumed.

- advanced_props_and_state_migrator

    Cases:
    1. Component already overrides `consumedProps`; do nothing
    1. Props class had no members, so no mixin was created: add `consumedProps` override that consumes no props
    1. Mixin was created from props/state class; add `consumedProps` override that consumes that mixin
    1. Props were moved from class into existing mixin; add `consumedProps` override that consumes that mixin and add a fixme

    Also, if any of these extend from a non-base component, add a fixme to ensure that that consumedProps impl shouldn't be inherited instead of overriding it in the base class

**2\. Update tests, which already had all of these cases covered, to match new behavior**

#### Release Notes
- Update boilerplate migrator to preserve consumedProps behavior

## Review
Please review: @aaronlademann-wf @joebingham-wk @sydneyjodon-wk 

### QA Checklist

## Merge Checklist

### DIFF
--- a/lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
+++ b/lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
@@ -18,6 +18,7 @@ import 'package:codemod/codemod.dart';
 import 'package:over_react_codemod/src/boilerplate_suggestors/migration_decision.dart';
 import 'package:source_span/source_span.dart';
 
+import '../constants.dart';
 import '../util.dart';
 import 'boilerplate_utilities.dart';
 
@@ -397,7 +398,6 @@ class AdvancedPropsAndStateClassMigrator extends GeneralizingAstVisitor
 
       // Consumed props changes don't apply to state classes
       if (isAPropsClass(node)) {
-
         // Update consumedProps to preserve behavior between legacy boilerplate
         // (only props in the main props class are consumed by default)
         // and the new boilerplate (all mixins are consumed by default).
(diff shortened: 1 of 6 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

test/boilerplate_suggestors/advanced_props_and_state_class_migrator_test.dart
```diff
@@ -387,7 +414,6 @@ void advancedPropsAndStateClassMigratorTestHelper({
               'ADifferentPropsClass': 'ADifferentPropsClassMixin',
             },
           )(
-            expectedPatchCount: 6,
```
greglittlefield-wf, line 390: I removed most of the `expectedPatchCount` assertions in this file, since they don't provide much value (patches are already tested via `expectedOutput`) and are a pain to update.

I left the ones where the value was `0`/1`.

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
```diff
@@ -392,6 +395,94 @@ class AdvancedPropsAndStateClassMigrator extends GeneralizingAstVisitor
...
+          {
+            final unprefixedSuperclassName = componentNode
+                .extendsClause?.superclass?.name?.name
+                ?.split('.')
+                ?.last;
+            final extendsFromOverReactComponentBaseClass =
+                overReactBaseComponentClasses
+                    .contains(unprefixedSuperclassName);
+
+            if (!extendsFromOverReactComponentBaseClass) {
```
sydneyjodon-wk, line 443: Should there be a null check for `unprefixedSuperclassName` here? Or would we still need a FIXME comment in that case?

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
```diff
@@ -392,6 +395,94 @@ class AdvancedPropsAndStateClassMigrator extends GeneralizingAstVisitor
...
+        final componentNode = getComponentNodeInRoot(node);
+        final consumedPropsDeclarations = componentNode.members
+            .whereType<MethodDeclaration>()
+            .where((method) => method.name.name == 'consumedProps');
+
+        // [1], [2]
+        if (!isAbstract(componentNode) && consumedPropsDeclarations.isEmpty) {
+          // [2]
+          var parentClassFixmeComment = '';
+          {
```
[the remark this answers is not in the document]
greglittlefield-wf: I actually did this on purpose so that `unprefixedSuperclassName` and `extendsFromOverReactComponentBaseClass` don't leak into the scope of the rest of the function; they're just intermediates in the computation of parentClassFixmeComment, which I wanted to keep self-contained

lib/src/boilerplate_suggestors/advanced_props_and_state_class_migrator.dart
greglittlefield-wf: Good question! 

I think it would be extremely unlikely for a component class to not have a superclass and make its way in here. And if that does happen, it'll just output `null` instead of crashing, which I think is good enough

greglittlefield-wf:
@Workiva/release-management-p

PR #97, opened 2020-08-11
CPLAT-11695 Add option to specify a path or glob upon which to run boilerplate upgrade codemod

smaifullerton-wk:
## Motivation
  
We direct consumers to re-run the boilerplate_upgrade script after making certain changes to individual components. However, there is no way for them to run the script on one or some files, only all files. Running the script on all files can lead to many changes in files other than the ones the consumer updated, which may cause him or her to abandon ship on the update entirely (✋ )

## Changes
  
This PR adds the ability to specify one or more paths or globs to the boilerplate_upgrade codemod to run the codemod only on the specified files.

#### Release Notes
  
- Add ability to pass paths or globs to boilerplate_upgrade as basic args to only run the codemod on the specified files.

## Review

  

Please review: 

### QA Checklist
- [ ] Run the `boilerplate_upgrade.dart` script locally
- [ ] Ensure that the `--help` flag prints help output
- [ ] Ensure that you can pass one or more paths or globs as a basic argument to the codemod, and that the codemod only runs on the files specified
- [ ] Ensure that not passing in a path as a basic arg runs the codemod on all of the Dart files where you run the codemod
- [ ] Ensure you can mix passing in paths/globs with other options for the codemod such as `--treat-all-components-as-private`

## Merge Checklist

### DIFF
--- a/lib/src/executables/boilerplate_upgrade.dart
+++ b/lib/src/executables/boilerplate_upgrade.dart
@@ -98,10 +98,7 @@ void main(List<String> args) {
   // If a path or multiple paths are provided as a basic arg, get file paths for them.
   Iterable<String> filePaths = [];
   if (pathArgs != null) {
-    Set<String> pathSet = {};
-    pathArgs.forEach((path) =>
-        pathSet.addAll(filePathsFromGlob(Glob(path, recursive: true))));
-    filePaths = pathSet;
+    filePaths = pathArgs.toSet();
     logger?.info("Codemod will run on these files: ${filePaths}");
   } else {
     filePaths = allDartPathsExceptHiddenAndGenerated();
--- a/lib/src/executables/boilerplate_upgrade.dart
+++ b/lib/src/executables/boilerplate_upgrade.dart
@@ -46,9 +46,9 @@ const _changesRequiredOutput = '''
   pub global activate over_react_codemod
   pub global run over_react_codemod:boilerplate_upgrade
 
-  Or you can use the -p flag to specify a path or glob to run the codemod on only some files:
-  pub global run over_react_codemod:boilerplate_upgrade -p "path/to/your/file.dart"
-  pub global run over_react_codemod:boilerplate_upgrade -p "lib/**.dart"
+  Or you can specify one or more paths or globs to run the codemod on only some files:
+  pub global run over_react_codemod:boilerplate_upgrade path/to/your/file.dart another/file.dart
+  pub global run over_react_codemod:boilerplate_upgrade lib/**.dart
   
   pub run dart_dev format (if you format this repository).
   Then, review the the changes, address any FIXMEs, and commit.
@@ -58,10 +58,6 @@ void main(List<String> args) {
   final parser = ArgParser()
     ..addFlag('help',
         abbr: 'h', negatable: false, help: 'Prints this help output')
-    ..addOption('path',
-        abbr: 'p',
-        help:
-            'Specify a path or glob to run codemod on just Dart files specified')
     ..addSeparator('Boilerplate Upgrade Options:')
     ..addFlag(_convertedClassesWithExternalSuperclassFlag,
         help: 'Converts classes with external superclasses')
@@ -85,7 +81,6 @@ void main(List<String> args) {
   }
 
   exitCode = upgradeReactVersions(
-    args: parsedArgs.rest,
     overReactVersionRange: parsedArgs[_overReactVersionRangeOption],
     overReactTestVersionRange: parsedArgs[_overReactTestVersionRangeOption],
   );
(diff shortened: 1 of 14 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/executables/boilerplate_upgrade.dart
```diff
@@ -76,20 +99,25 @@ void main(List<String> args) {
...
 
-  final query = FileQuery.dir(
-    pathFilter: (path) {
-      return isDartFile(path) && !isGeneratedDartFile(path);
-    },
-    recursive: true,
-  );
+  final path = parsedArgs['path'];
+  // If a path or glob is provided as an arg, get file paths for just that path or glob.
+  Iterable<String> fileQuery = path != null
```
smaifullerton-wk, line 104: I spent >1 hour trying to test this by adding a new unit test suite that runs the `boilerplate_upgrade` executable, but ran into issues getting past the first `✔ Are these ranges correct? (Y/n)` message using `Process.start` and `stdin`.
There aren't any existing unit tests for this executable, so I decided it wasn't that important, but if you feel it's critical to test this, I can try again.

lib/src/executables/boilerplate_upgrade.dart
```diff
@@ -44,23 +45,45 @@ const _changesRequiredOutput = '''
...
 
   final logger = Logger('over_react_codemod.boilerplate_upgrade');
 
+  if (parsedArgs['help'] == true) {
+    stderr.writeln(parser.usage);
+    return;
+  }
+
   exitCode = upgradeReactVersions(
     args: parsedArgs.rest,
```
evanweible-wf, line 88: This `args` param gets passed on to `runInteractiveCodemod()`'s `args` param, which is intended to make it possible for consumers to setup a codemod executable without having to do any of its own arg parsing, e.g.:

```dart
main(List<String> args) {
  exitCode = runInteractiveCodemod(..., args: args);
}
```

But in this case, the codemods have their own arg parsing implemented, which means that any flags/options not explicitly added to the above parser are unsupported. For example, a consumer of this codemod cannot pass in the `--yes-to-all` flag that is supported by `runInteractiveCodemod` because the above call to `parser.parse(args)` would throw when it is present. If we wanted to support these options that `runInteractiveCodemod` supports, we'd have to add them to the arg parser used above (or maybe export `codemodArgParser` from `package:codemod` so that it can be extended for use cases like this).

In other words, passing `parsedArgs.rest` here isn't doing anything.

---

All of that is to say, I think it'd be better to use `parsedArgs.rest` as a way to allow consumers to specify exactly which files to run the codemod on instead of the `-p|--path` option, since that's what the basic args are traditionally used for and it allows consumers to utilize their own shell's expansion to do something like this:

```shell
pub global run over_react_codemod:boilerplate_upgrade lib/**.dart
```

This also simplifies the implementation here slightly because you can treat the args as a simple list of file paths.

smaifullerton-wk: Thanks for this. I pushed a commit to change the pattern to use basic args instead, and removed the doing-nothing `parsedArgs.rest` params.

I'm not too concerned about addressing passing dart-codemod-specific options through correctly to `runInteractiveCodemod` -- it's never come up, and the vast majority of the usage of this codemod is behind us.

lib/src/executables/boilerplate_upgrade.dart
```diff
@@ -76,20 +94,28 @@ void main(List<String> args) {
...
-    recursive: true,
-  );
+  final pathArgs = parsedArgs.rest.isNotEmpty ? parsedArgs.rest : null;
+  // If a path or multiple paths are provided as a basic arg, get file paths for them.
+  Iterable<String> filePaths = [];
+  if (pathArgs != null) {
+    Set<String> pathSet = {};
+    pathArgs.forEach((path) =>
+        pathSet.addAll(filePathsFromGlob(Glob(path, recursive: true))));
+    filePaths = pathSet;
```
evanweible-wf, line 104: You shouldn't need to create and resolve `Glob`s from the input paths - shell expansion happens before the executable is run, so something like this:

```
pub global run over_react_codemod:boilerplate_upgrade lib/**.dart
```

would result in `parsedArgs.rest` being a list of all the file paths within `lib/`, not the original `lib/**.dart` pattern.

The only downside with this approach that I'm realizing now is that `runInteractiveCodemod()` expects file paths, so if someone were to pass in a directory like `lib/` as a path, that would result in an error. We should probably have `runInteractiveCodemod` expand directory paths to include all of the files within that directory

evanweible-wf: I take that last part back. If `runInteractiveCodemod` does its own expanding of directories, then we're right back where we were before with not really knowing which files to include by default.. e.g. if someone passes in `lib/` and we expand that, there might be SASS files or images in that directory that make no sense to run the codemod on. Leaving it up to the consumer to use an appropriate glob like `lib/**.dart` seems sufficient to me in this case.

smaifullerton-wk: Addressed!

lib/src/executables/boilerplate_upgrade.dart
```diff
@@ -124,8 +147,8 @@ void main(List<String> args) {
       GeneratedPartDirectiveIgnoreRemover(),
       FactoryIgnoreCommentMover(),
     ].map((s) => Ignoreable(s)),
-    args: parsedArgs.rest,
     defaultYes: true,
+    args: [],
```
corwinsheahan-wf, line 151: Can we not pass anything for `args`, or do we need an empty list?

lib/src/executables/boilerplate_upgrade.dart
```diff
@@ -17,6 +17,7 @@ import 'dart:io';
 import 'package:args/args.dart';
 import 'package:codemod/codemod.dart';
 import 'package:collection/collection.dart';
+import 'package:glob/glob.dart';
```
corwinsheahan-wf, line 20: Is this import needed?

lib/src/executables/boilerplate_upgrade.dart
smaifullerton-wk: We need an empty list. Null throws

sydneyjodon-wk:
QA in progress

smaifullerton-wk:
@Workiva/release-management-p

PR #106, opened 2020-12-11
CPLAT-12621 Codemod to update factory syntax for Dart 2.9

sydneyjodon-wk:
## Motivation
  
We need a codemod to update factory syntax to the new Dart 2.9 boilerplate.

Factory Declarations:
```diff
- UiFactory<FooProps> Foo = _$Foo; // ignore: undefined_identifier, invalid_assignment
+ UiFactory<FooProps> Foo = castUiFactory(_$Foo); // ignore: undefined_identifier
```

Function Components:
```diff
UiFactory<FooProps> Foo = uiFunction(
    (props) {},
-   $FooConfig, // ignore: undefined_identifier, argument_type_not_assignable
+   _$FooConfig, // ignore: undefined_identifier
);
```

Connected Components:
```diff
UiFactory<FooProps> Foo =
    connect<SomeState, FooProps>(
  mapStateToProps: (state) => (Foo()
    ..foo = state.foo
    ..bar = state.bar
  ),
- )(_$Foo); // ignore: undefined_identifier, argument_type_not_assignable
+ )(castUiFactory(_$Foo); // ignore: undefined_identifier
```

## Changes
  
- Created `dart2_9_upgrade` codemod executable to bump the `over_react` lower bound and run the following suggestors:
  - `FactoryAndConfigIgnoreCommentRemover`: to remove ignore comments (particularly `argument_type_not_assignable` and `invalid_assignment`).
  - `FactoryConfigMigrator`: to add the private prefix to generated factory configs.
  - `GeneratedFactoryMigrator`: to wrap generated factories in the `castUiFactory` function.
- Wrote tests.

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
      - [ ] CI passes
      - [ ] Verify that tests cover all cases
      - [ ] Pull down this branch and:
        - [ ] Run `pub global activate -s path .` in the codemod repo
        - [ ] Then run `pub global run over_react_codemod:dart2_9_upgrade` in multiple repos and verify that it works as expected.
          - [ ] Verify that all connected components, component factory declarations, function components (and `uiForwardRef`) usages are updated correctly.
          - [ ] Verify that codemod runs without errors even when parsing invalid code (specifically in the `over_react` analyzer plugin playground).
          - [ ] Verify that the codemod makes no additional changes the second time it is run.
          - [ ] Verify that factory declarations in the legacy boilerplate syntax are not updated.
          - [ ] Verify that `argument_type_not_assignable` and `invalid_assignment` ignore comments are removed (and other ignore comments are not)
            - To more thoroughly QA `FactoryAndConfigIgnoreCommentRemover`, try adding `FactoryAndConfigIgnoreCommentRemover('undefined_identifier')` [here](https://github.com/Workiva/over_react_codemod/blob/e7c6ae082db31d02f9094c1cf6f961d64ade5ad8/lib/src/executables/dart2_9_upgrade.dart#L55-L56) and verify that all `undefined_identifier` ignore comments are removed (regardless of their location around the generated factory).

## Merge Checklist


aviary-wf:
## Security Insights

## Action Items
---

lib/src/dart2_9_suggestors/dart2_9_utilities.dart
```diff
@@ -0,0 +1,164 @@
...
+/// | (Foo)                          | null   |
+SimpleIdentifier getGeneratedFactoryArg(ArgumentList argList) {
+  final args = argList.arguments;
+  if (args.length != 1) return null;
+
+  final generatedPrefix = '_\$';
+
+  final method =
+      argList.thisOrAncestorOfType<FunctionExpressionInvocation>()?.function;
+  if (method is MethodInvocation && method.methodName.name == 'connect') {
```
aaronlademann-wf, line 42: So `getGeneratedFactoryArg` is a function intended only for `connect`? Can we update the name / description to indicate that?

lib/src/dart2_9_suggestors/dart2_9_utilities.dart
```diff
@@ -0,0 +1,164 @@
...
+/// Iterates through the [comment] token stream and removes all ignore
+/// comments containing [ignoreToRemove] using the [yieldPatch] provided.
+///
+/// The entire ignore comment will only be removed if [ignoreToRemove] is the only
+/// item in the list to ignore. Otherwise, [ignoreToRemove] will just be
+/// removed from the list.
+///
+/// Example:
+///
+/// Calling:
```
joebingham-wk, line 122: ~Are there meant to be examples between `Example` and `Calling`?~

Ohh I get it, nevermind. #nit I think the wording could be clarified or could use indentation, but I caught up! 😄

lib/src/dart2_9_suggestors/dart2_9_utilities.dart
sydneyjodon-wk: Yep, you're right! I'll update that

test/dart2_9_suggestors/dart2_9_utilities_test.dart
```diff
@@ -0,0 +1,410 @@
...
+          _expectConfigName(
+            input: '''
+              UiFactory<FooProps> Foo = uiForwardRef(
+                (props) {},
+                _\$FooConfig, // ignore: undefined_identifier
+              );
+            ''',
+            expectedName: '_\$FooConfig',
+          );
+        });
```
joebingham-wk, line 223: Could we get a test here showing a `uiFunction` wrapped in `memo`?

test/dart2_9_suggestors/factory_and_config_ignore_comment_remover_test.dart
```diff
@@ -0,0 +1,553 @@
...
+              _\$Foo; // ignore: unused_element
+          ''',
+        );
+      });
+
+      test('from a non-component factory declaration', () {
+        testSuggestor(
+          expectedPatchCount: 0,
+          input: r'''
+            DriverFactory driverFactory = createDriver; // ignore: undefined_identifier
```
joebingham-wk, line 83: Should this be:
```suggestion
            DriverFactory driverFactory = createDriver; // ignore: $ignoreToRemove
```

joebingham-wk (commented):
This looks awesome! I caught a couple small things, but yeah, pretty much everything was really clear and looked great!

test/dart2_9_suggestors/factory_and_config_ignore_comment_remover_test.dart
sydneyjodon-wk: oh yep! Good catch!

lib/src/dart2_9_suggestors/dart2_9_utilities.dart
sydneyjodon-wk: Updated in 17b25b1

test/dart2_9_suggestors/dart2_9_utilities_test.dart
sydneyjodon-wk: Added in 17b25b1

lib/src/dart2_9_suggestors/factory_and_config_ignore_comment_remover.dart
```diff
@@ -0,0 +1,161 @@
...
+  @override
+  visitArgumentList(ArgumentList node) {
+    super.visitArgumentList(node);
+
+    final generatedArg = getGeneratedFactoryConfigArg(node) ??
+        getConnectGeneratedFactoryArg(node);
+    if (generatedArg != null) {
+      final commentList = List<Token>();
+
+      // Check comments before the generated argument.
```
greglittlefield-wf, line 46: Instead of enumerating each case and potentially missing some, a more generic check could be to look at line numbers to tell whether comments apply.

```dart
      final generatedArgLineNumber = sourceFile.getLine(generatedArg.offset);
      for (final comment in allComments(node.root.beginToken)) {
        final commentLineNumber = sourceFile.getLine(comment.offset);
        final commentAppliesToGeneratedArg =
            // EOL comments 
            commentLineNumber == generatedArgLineNumber ||
            // Comments on the previous line
            commentLineNumber == generatedArgLineNumber - 1;
        if (commentAppliesToGeneratedArg) {
          commentList.add(comment);
        }
      }
```

With that implementation, I think you'd also want to remove the recursive logic from `removeIgnoreComment`, since you're already hitting all comments.

lib/src/dart2_9_suggestors/dart2_9_utilities.dart
```diff
@@ -0,0 +1,170 @@
...
+  if (type != null && type is NamedType && type?.name?.name == 'UiFactory') {
+    final initializer = node.variables?.variables?.first?.initializer;
+    if (initializer is SimpleIdentifier) {
+      return initializer.name.startsWith('_\$');
+    } else if (initializer is MethodInvocation &&
+        initializer.methodName.name == castFunctionName) {
+      return initializer.argumentList.arguments.first
+          .toSource()
+          .startsWith('_\$');
+    }
```
greglittlefield-wf, line 105: This could be simplified, and cover more cases (such as `connect` and wrapper functions, as well as deeply nested wrapper functions), by checking to see if the generated factory is used anywhere within the initializer (which is the same as what we do in the builder):
```suggestion
    if (initializer != null) {
      final name = node.variables.variables.first.name.name;
      final generatedName = r'_$' + name;
      return allDescendantsOfType<SimpleIdentifier>(initializer)
          .any((identifier) => identifier.name == generatedName);
    }
```
(using allDescendantsOfType [from over_react](https://github.com/Workiva/over_react/blob/ac85e9d2fdbd7fa7f9f0174c845e3c8951b46e95/tools/analyzer_plugin/lib/src/util/ast_util.dart#L40-L57))

A similar approach could be taken to find the identifier where `getConnectGeneratedFactoryArg` is used, allowing you to cover both the basic and `connect` cases with the same code.

I think this improvement would be worth it, but feel free to disagree.

lib/src/dart2_9_suggestors/dart2_9_utilities.dart
```diff
@@ -0,0 +1,170 @@
...
+///   `// ignore: invalid_assignment` => `// ignore: invalid_assignment`
+void removeIgnoreComment(
+    Token comment, String ignoreToRemove, YieldPatch yieldPatch) {
+  if (comment == null) return;
+
+  final lexeme = comment.lexeme.replaceAll(' ', '').toLowerCase();
+  if (lexeme.startsWith('//ignore:')) {
+    final ignoreList =
+        lexeme.replaceFirst(RegExp('\/\/ignore\:'), '').split(',');
+    if (ignoreToRemove == null ||
```
greglittlefield-wf, line 151: It looks like the comment is always removed if ignoreToRemove is `null`.

Is that used anywhere? If so, it should be documented. 

If not, perhaps that functionality should be removed and a Argument.notNull error thrown.

lib/src/util.dart
```diff
@@ -611,3 +612,13 @@ Iterable<String> allDartPathsExceptHidden() =>
 Iterable<String> allDartPathsExceptHiddenAndGenerated() =>
     filePathsFromGlob(Glob('**.dart', recursive: true))
         .where((path) => !path.endsWith('.g.dart'));
+
+/// Returns whether or not [sourceText] contains invalid code and, therefore,
+/// should be skipped when generating patches.
+///
+/// Used to override [Suggestor.shouldSkip] when parsing errors should be
+/// skipped instead of thrown when [Suggestor.generatePatches] is called.
+bool shouldSkipParsingErrors(String sourceText) {
```
greglittlefield-wf, line 621: This name might imply that it's making a determination, based on the contents of a file, whether it's okay to skip parsing errors.

I think a better name for this function would be
```suggestion
bool hasParseErrors(String sourceText) {
```
which more clearly states what this function is doing, and also makes it clearer at the call-sites why exactly we're skipping a file.

```dart
bool shouldSkip(String sourceText) => hasParseErrors(sourceText);
```

test/dart2_9_suggestors/factory_and_config_ignore_comment_remover_test.dart
```diff
@@ -0,0 +1,553 @@
...
+            ''',
+          );
+        });
+      });
+
+      test('when the factory is already updated', () {
+        testSuggestor(
+          expectedPatchCount: 1,
+          input: '''
+            UiFactory<FooProps> Foo = $castFunctionName(_\$Foo); // ignore: $ignoreToRemove
```
greglittlefield-wf, line 216: Shouldn't the codemod keep `undefined_identifier` ignores in place for Dart <2.12 compatibility?

greglittlefield-wf: Oh wait, is it because you're calling `ignoreRemoverTestHelper('undefined_identifier');` in the test, but not passing in `'undefined_identifier'` in the real codemod?

greglittlefield-wf (dismissed):

lib/src/dart2_9_suggestors/dart2_9_utilities.dart
sydneyjodon-wk: oh good catch! That definitely should not happen. I'll add `ArgumentError.checkNotNull`

test/dart2_9_suggestors/factory_and_config_ignore_comment_remover_test.dart
sydneyjodon-wk: Yep! I just tested the suggestor using `undefined_identifier` because that was the case where the ignore comments would be in all those locations in real life, but I also tested `ignoreRemoverTestHelper('argument_type_not_assignable')` to verify that it removes whichever comment is given in the constructor. But, yes, the real codemod is [only removing](https://github.com/Workiva/over_react_codemod/blob/17b25b1ee71fc4e27fed47d9ebca3e593ef36e64/lib/src/executables/dart2_9_upgrade.dart#L55-L56) `invalid_assignment` and `argument_type_not_assignable` right now.

lib/src/dart2_9_suggestors/factory_and_config_ignore_comment_remover.dart
sydneyjodon-wk: That makes it so much nicer! Updated in ebecf54

lib/src/dart2_9_suggestors/dart2_9_utilities.dart
sydneyjodon-wk: That does look nicer! Right now over_react_codemod doesn't have over_react as a dependency - is it worth adding the dependency for one utility? 

Also is `allDescendantsOfType` exported anywhere in over_react? I can't seem to access it, but I might be missing something.

greglittlefield-wf: Oh sorry, I should have clarified; I'd recommend just copy-pasting the code from that repo, since it's such a small implementation.

sydneyjodon-wk: Ohh 👍

lib/src/dart2_9_suggestors/generated_factory_migrator.dart
```diff
@@ -25,37 +25,26 @@ import 'dart2_9_utilities.dart';
...
-        yieldPatch(initializer.end, initializer.end, ')');
-      }
+    if (isClassOrConnectedComponentFactory(node) &&
+        !isLegacyFactoryDecl(node)) {
+      final generatedFactory = getGeneratedFactory(node);
+      if (generatedFactory
+              .thisOrAncestorOfType<MethodInvocation>()
+              ?.methodName
+              ?.name ==
+          castFunctionName) return;
```
greglittlefield-wf, line 40: Due to formatting this should use curly-braces

lib/src/dart2_9_suggestors/generated_factory_migrator.dart
```diff
@@ -25,37 +25,26 @@ import 'dart2_9_utilities.dart';
...
-      if (initializer is SimpleIdentifier) {
-        yieldPatch(
-            initializer.offset, initializer.offset, '$castFunctionName(');
-        yieldPatch(initializer.end, initializer.end, ')');
-      }
+    if (isClassOrConnectedComponentFactory(node) &&
+        !isLegacyFactoryDecl(node)) {
+      final generatedFactory = getGeneratedFactory(node);
+      if (generatedFactory
+              .thisOrAncestorOfType<MethodInvocation>()
```
greglittlefield-wf, line 37: #nit This checks for any ancestor, but we might want to make sure it's wrapped directly around the generated factory:

```suggestion
      final generatedFactory = getGeneratedFactory(node);
      final parent = generatedFactory.parent;
      if (parent is MethodInvocation && parent
```

lib/src/util.dart
```diff
@@ -49,6 +52,12 @@ Iterable<Token> allComments(Token beginToken) sync* {
     }
     currentToken = currentToken.next;
   }
+  // Also check comments preceding EOF.
```
greglittlefield-wf, line 55: Ohh good catch. 

#nit Instead of adding a loop for this case, could we just update the while-loop to remove the `isEof` check, since the EOF token should have a null `.next`?
```diff
-while (currentToken != null && !currentToken.isEof) {
+ hile (currentToken != null) {
```

test/util_test.dart
```diff
@@ -20,6 +20,7 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
 import 'package:pub_semver/pub_semver.dart';
+import 'package:react/react.dart';
```
greglittlefield-wf, line 23: is this import necessary? I would expect it to cause issues since this is a VM test suite

greglittlefield-wf (changes requested):
Just a few more comments and some nits, this looks great!

lib/src/util.dart
sydneyjodon-wk: I tried that, but for some reason `.next` of the EOF token is itself so it made an infinite loop 🤷‍♀️

test/util_test.dart
sydneyjodon-wk: oh nope. I forgot to check for lints 🤦‍♀️

lib/src/util.dart
greglittlefield-wf: Ah, gotcha, weeeird 👍

greglittlefield-wf (approved):

joebingham-wk (commented):
QA +1! Ran the codemod in over_react and audit and checked the following:

- [x] Then run `pub global run over_react_codemod:dart2_9_upgrade` in multiple repos and verify that it works as expected.
  - [x] Verify that all connected components, component factory declarations, function components (and `uiForwardRef`) usages are updated correctly.
  - [x] Verify that codemod runs without errors even when parsing invalid code (specifically in the `over_react` analyzer plugin playground).
  - [x] Verify that the codemod makes no additional changes the second time it is run.
  - [x] Verify that factory declarations in the legacy boilerplate syntax are not updated.
  - [x] Verify that `argument_type_not_assignable` and `invalid_assignment` ignore comments are removed (and other ignore comments are not)
    - [x] To more thoroughly QA `FactoryAndConfigIgnoreCommentRemover`, try adding `FactoryAndConfigIgnoreCommentRemover('undefined_identifier')` [here](https://github.com/Workiva/over_react_codemod/blob/e7c6ae082db31d02f9094c1cf6f961d64ade5ad8/lib/src/executables/dart2_9_upgrade.dart#L55-L56) and verify that all `undefined_identifier` ignore comments are removed (regardless of their location around the generated factory).

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #129, opened 2021-10-11
CPLAT-15826 Codemod to add `ThemeProvider` to `react_dom.render`

sydneyjodon-wk:
## Motivation
  
In order to allow RMUI components to be used anywhere in Wdesk, we need to wrap all production and example `react_dom.render` calls in `ThemeProvider`.

This needs to be done before migrations take place. Otherwise, packages that perform their own react_dom.render calls could consume UI from other packages that have been migrated to RMUI without a ThemeProvider, resulting in the default MUI theming being displayed.

## Changes
  
- Add a `wk_theme_provider` codemod that wraps render calls in `ThemeProvider()..theme = wkTheme` and adds the `theme_provider.dart` import.
- Write tests.

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
      - [ ] CI passing and good test coverage
      - [ ] Pull down this branch and run: `pub global activate -s path .`
        - [ ] Run the codemod on another repo and verify that it works as expected:
          - `pub global run over_react_codemod:rmui_preparation --yes-to-all` 
          - Make sure `ThemeProvider`s are added inside `ErrorBoundary`s if they exist
          - Make sure render calls in test directories are not updated

## Merge Checklist


sydneyjodon-wk:
@aaronlademann-wf There's not a benefit to having it separate - I went back and forth on it and ended up making it separate just because of naming since it's not "bundle" related and since the `dependency_validator` ignore part ended up being a separate codemod, I didn't know if we were trying to keep things separate.

But I would be fine with consolidating the codemods into one - maybe even making the dependency_validator one a part of the same codemod? What do you think?

greglittlefield-wf:
I believe these are unresolved AST suggestors (`AstVisitingSuggestor.shouldResolveAst` defaults to `false`), so in this case it wouldn't make a noticeable performance difference.

For the component codemods, we do end up having to resolve the AST more than once so that we can pick up on changes applied from previous suggestors, and while it's a bit of a hit, it's not terrible (I've seen wdesk_sdk take ~2.5 minutes to resolve once and an additional ~30 seconds to resolve a second time).

lib/src/theme_provider_suggestors/theme_provider_adder.dart
```diff
@@ -0,0 +1,121 @@
...
+
+  ThemeProviderAdder(this.theme);
+
+  @override
+  visitMethodInvocation(MethodInvocation node) {
+    super.visitMethodInvocation(node);
+
+    // Don't update if [node] is not a render call or the file is in a test directory.
+    if (node.methodName.name != 'render' ||
+        context.relativePath.split('/').contains('test')) return;
```
greglittlefield-wf, line 35: Note that this check could potentially skip over [subdirectories named `test` not in the top-level `test` directory](https://sourcegraph.wk-dev.wdesk.org/search?q=file:lib%5Cb.*/test/+lang:dart+-repo:cp_labs+-repo:messaging-sdk+-repo:puppeteer-platform+-repo:frugal&patternType=literal).

From the linked usages, this shouldn't be an issue, but I think it'd be worth calling out in a code comment

greglittlefield-wf (changes requested):

test/theme_provider_suggestors/theme_provider_adder_test.dart
```diff
@@ -0,0 +1,597 @@
...
+            return react_dom.render((Foo()..color = 'red')(), mountNode);
+          }
+        ''',
+        expectedOutput: '''
+          import 'package:over_react/over_react.dart';
+          import 'package:react/react_dom.dart' as react_dom;
+          import 'package:react_material_ui/styles/theme_provider.dart';
+
+          main() {
+            return react_dom.render((ThemeProvider()..theme = wkTheme)((Foo()..color = 'red')()), mountNode);
```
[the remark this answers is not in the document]
sydneyjodon-wk: @greglittlefield-wf do we also want a fixme for variable assignments?

lib/src/theme_provider_suggestors/theme_provider_adder.dart
```diff
@@ -106,16 +119,19 @@ class ThemeProviderAdder extends GeneralizingAstVisitor
...
-        patch,
-        renderFirstArg.offset,
-        renderFirstArg.offset,
-      );
+      final parent = node.parent;
+      if (parent is ArgumentList ||
+          parent is VariableDeclaration ||
+          parent is AssignmentExpression ||
+          parent is ReturnStatement ||
+          parent is ExpressionFunctionBody) {
```
greglittlefield-wf, line 127: Since expressions can be nested within many types of nodes, a more robust way to handle this would be to instead check for the the cases we're sure we don't need to check the result of the expression. Which, I think is only a single case: an expression by itself as a statement:
```dart
someFunction() {
  react_dom.render(...);
}
```

which we could handle with
```dart
if (parent is! ExpressionStatement) {
  // Add a comment to verify the typing if the return value
  // of react_dom.render could be used.
   yieldPatch(' $checkTypingFixmeComment\n', node.offset, node.offset);
}
```

lib/src/theme_provider_suggestors/theme_provider_adder.dart
```diff
@@ -18,6 +18,9 @@ import 'package:codemod/codemod.dart';
 import 'package:collection/collection.dart' show IterableExtension;
 import 'package:over_react_codemod/src/util/component_usage.dart';
 
+const checkTypingFixmeComment =
+    '// FIXME: Verify this typing now that it is wrapped in a ThemeProvider.';
```
greglittlefield-wf, line 22: This comment is a little vague, and also I think we need to review more than just the typing, since ThemeProvider doesn't have a component instance.

It should be something along the lines of "Verify the return value of react_dom.render is unused, since the wrapper ThemeProvider has no component instance. If a component instance is needed, set a ref on the child component." (potentially worded more concisely 😄 )

aaronlademann-wf (approved):
QA +1

- Tested in `w_table`
- All `react_dom.render` calls now contain the `ThemeProvider` with `props.theme` set to `wkTheme` within existing `ErrorBoundary` components.

PR #136, opened 2021-11-16
CPLAT-16174 Fix dynamic calls involving toBuilder() being detected as unresolved usages

greglittlefield-wf:
## Motivation
There are some usages of built_value's `Built.toBuilder` method that cause issues detected as component usages.

Take, for example, the expression in the following function:
```dart
void example(Built b) {
  b.toBuilder();
}
```
When the `toBuilder` usages are resolved, the `getComponentUsage` is usually able to use the static type information available to determine that they're not OverReact component usages, since the result of the `toBuilder()` call isn't a subtype of `UiProps`, so `getComponentUsage` returns `null`.

However, if toBuilder is called on a `dynamic` object, such as in this example:
```dart
void example(dynamic b) {
  b.toBuilder();
}
```
then `getComponentUsage` doesn't have any static information on what type `toBuilder` returns (similar to if this call wasn't fully resolved due to some analysis issue), so it falls back to using the name of the method to determine if it looks like a component builder (by seeing if the name contains the string "builder").

## Changes
I started trying to detect dynamic calls and ignore those cases, but unfortunately, it's pretty difficult to tell the difference between members accessed on dynamic expressions and an unresolved member accesses.

So, I opted to specifically check for methods named `toBuilder` (only when the AST is unresolved) and just exclude them as a workaround.

I also added tests, and updated a helper method to reflect the gotcha with dynamic member accesses I mentioned above.

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        - Verify that running `mui_migration` with this branch of the codemod activated against workflow_client works without throwing errors
            - To activate this branch globally, check out the branch, and within `over_react_codemod` run `pub global activate -s path .`

## Merge Checklist

### DIFF
--- a/lib/src/util/component_usage.dart
+++ b/lib/src/util/component_usage.dart
@@ -48,7 +48,7 @@ class FluentComponentUsage {
 
   /// Whether this usage's builder and all descendant expressions (including the
   /// factory, if it exists) have fully resolved static information available.
-  bool get isBuilderResolved => isFullyResolved(builder);
+  bool get isBuilderResolved => !hasUnresolvedOrDynamicReferences(builder);
 
   /// The top-level variable element for this usage's factory, or `null` if either:
   ///
@@ -75,7 +75,7 @@ class FluentComponentUsage {
 
   /// The static type of this usage's builder, or `null` if this usage is not fully resolved.
   DartType? get builderType {
-    if (isFullyResolved(builder)) {
+    if (!hasUnresolvedOrDynamicReferences(builder)) {
       return builder.staticType;
     }
     return null;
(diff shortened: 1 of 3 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/util/component_usage.dart
```diff
@@ -495,29 +495,35 @@ class IndexPropAssignment extends BuilderMemberAccess {
...
+/// 1. are not fully resolved
+/// 2. represent members accessed on `dynamic` expressions (e.g., `dynamicVar.foo`)
+///
+/// Unfortunately, it's difficult to distinguish between the two and only detect
+/// unresolved references.
+///
+/// If dynamic member accesses are rare, this can a more accurate alternative
+/// than just checking for [Expression.staticType], since that often shows up
+/// as `dynamic` in resolved contexts when something doesn't fully resolve.
+bool hasUnresolvedOrDynamicReferences(Expression expression) {
```
greglittlefield-wf, line 510: This method and related code were updated to return the boolean inverse of what it used to return due to the name being updated.

lib/src/util/component_usage.dart
```diff
@@ -564,29 +570,51 @@ FluentComponentUsage? getComponentUsage(InvocationExpression node) {
...
   // but something goes wrong, we'll get dynamic.
-  if (isFullyResolved(builder)) {
+  if (!hasUnresolvedOrDynamicReferences(builder)) {
     // Resolved AST
     isComponent = builder.staticType!.isPropsClass;
   } else {
     // Unresolved AST (or type wasn't available)
-    isComponent = false;
 
     if (builder is MethodInvocation) {
```
greglittlefield-wf, line 591: These changes are better viewed with [whitespace-agnostic diff](https://github.com/Workiva/over_react_codemod/pull/136/files?w=1#diff-47ce059a0783e909caf8687c0b06e699647573ce171bcc8d6e4812375c4cd665R591)

lib/src/util/component_usage.dart
```diff
@@ -564,29 +570,51 @@ FluentComponentUsage? getComponentUsage(InvocationExpression node) {
     builder = functionExpression;
   }
 
-  bool isComponent;
```
greglittlefield-wf, line 567: I changed this variable to `final` so that the analyzer would enforce that it's assigned exactly one time (and isn't ever uninitialized or overwritten), which I felt made it easier to ensure there weren't any unhandled cases in the relatively complex if-else conditions below.

test/util/component_usage_migrator_test.dart
```diff
@@ -24,6 +24,10 @@ final sharedContext = SharedAnalysisContext.overReact;
 
 main() {
   group('ComponentUsageMigrator', () {
+    setUpAll(() async {
+      await SharedAnalysisContext.overReact.warmUpAnalysis();
+    });
```
greglittlefield-wf, line 29: CI was being flaky with timeouts, and adding this setUpAll (which is present in most other test files) fixed it!

joebingham-wk (dismissed):
+10! This looks great!
- [x] Ran this branch on workflow_client and verified it fixed the issue

aaronlademann-wf (approved):
+1

rmconsole-wf (approved):
+1 from RM

PR #142, opened 2022-04-04
INTL-2 Intl Migrator

matthewward-wf:
## Motivation
  
A set of migrators used for wrapping user facing strings in `dart:intl`'s `Intl.message` function
## Changes
  
Added migrators to help automate this process.
## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/executables/intl_message_migration.dart
```diff
@@ -0,0 +1,315 @@
...
+      [displayNameMigrator],
+      [importMigrator]
+    ]);
+
+    processedPackages.add(package);
+    if (exitCode != 0) {
+      outputFile.deleteSync();
+    } else {
+      List<String> lines = outputFile.readAsLinesSync();
+      if (lines.length == 4) {
```
greglittlefield-wf, line 229: Where does this 4 come from?

lib/src/intl_suggestors/utils.dart
```diff
@@ -0,0 +1,346 @@
...
+
+import 'constants.dart';
+
+// -- Node Validation --
+bool isValidStringInterpolationNode(AstNode node) {
+  if (!(node is StringInterpolation)) return false;
+  //We do not need to localize single values.  This should be handled by the
+  // variable being passed in.  IE: ..foo = '$bar' becomes [', '$bar', ']
+  if (node.elements.length == 3 &&
+      node.elements.first.toString() == node.elements.last.toString())
```
greglittlefield-wf, line 17: It seems like this case should be also checking that the elements are empty strings

lib/src/intl_suggestors/utils.dart
```diff
@@ -0,0 +1,346 @@
...
+import 'package:analyzer/dart/element/type.dart';
+import 'package:file/file.dart';
+import 'package:over_react_codemod/src/util/component_usage.dart';
+import 'package:collection/collection.dart' show IterableExtension;
+
+import 'constants.dart';
+
+// -- Node Validation --
+bool isValidStringInterpolationNode(AstNode node) {
+  if (!(node is StringInterpolation)) return false;
```
greglittlefield-wf, line 13: FYI you can use `is!` for these checks
```suggestion
  if (node is! StringInterpolation) return false;
```

lib/src/intl_suggestors/utils.dart
```diff
@@ -0,0 +1,346 @@
...
+  if (node.elements.length == 3 &&
+      node.elements.first.toString() == node.elements.last.toString())
+    return false;
+  return true;
+}
+
+bool isValidStringLiteralNode(AstNode node) {
+  if (!(node is SimpleStringLiteral)) return false;
+  if (node.value.isEmpty) return false;
+  if (double.tryParse(node.stringValue!) != null) return false;
```
greglittlefield-wf, line 25: #nit there's a non-nullable getter for this
```suggestion
  if (double.tryParse(node.value) != null) return false;
```

lib/src/intl_suggestors/utils.dart
```diff
@@ -0,0 +1,346 @@
...
+  } else {
+    return null;
+  }
+}
+
+bool excludeUnlikelyExpressions<E extends Expression>(PropAssignment prop,
+    String propKey) {
+  if (prop.rightHandSide.staticType == null) return true;
+  if (prop.rightHandSide.staticType?.isDartCoreBool ?? false) return true;
+  if (prop.rightHandSide.staticType?.isDartCoreNull ?? false) return true;
```
greglittlefield-wf, line 250: FYI if you assign to a local variable, the type can be promoted to non-nullable (or subtypes) based on if-checks.
```suggestion
  final staticType = prop.rightHandSide.staticType;
  if (staticType == null) return true;
  if (staticType.isDartCoreBool) return true;
  if (staticType.isDartCoreNull) return true;
```

lib/src/intl_suggestors/utils.dart
```diff
@@ -0,0 +1,346 @@
...
+      'Iterable<ReactElement>') return true;
+  final source = prop.rightHandSide.toSource();
+  if (source == propKey ||
+      source == 'props.$propKey' ||
+      source == "props['$propKey']") return true;
+  if (source == "''") return true;
+  if (source == "'.'") return true;
+  if (source == "'('") return true;
+  if (source == "')'") return true;
+  // If a string value is wrapped in quotes, and is in lowerCamelCase or UpperCamelCase, it is most likely a key of some kind, not a human-readable, translatable string.
```
greglittlefield-wf, line 265: Seems like this could reuse the logic in `quotedCamelCase`

lib/src/intl_suggestors/intl_migrator.dart
```diff
@@ -0,0 +1,118 @@
...
+          isValidStringLiteralNode(child.node) ||
+          isValidStringInterpolationNode(child.node));
+
+  @override
+  void migrateUsage(FluentComponentUsage usage) {
+    super.migrateUsage(usage);
+    final namePrefix =
+        (usage.node.thisOrAncestorOfType<ClassDeclaration>()?.name ??
+                usage.node.thisOrAncestorOfType<VariableDeclaration>()?.name)
+            .toString();
```
greglittlefield-wf, line 44: I think it'd be clearer here if the null case wasn't handled using `toString`
```suggestion
            ?.name ?? 'null';
```

lib/src/intl_suggestors/utils.dart
```diff
@@ -0,0 +1,346 @@
...
+String toClassName(String str) {
+  final res = toAlphaNumeric(str.replaceAll('_', ' '));
+  final pascalString = res
+      .splitMapJoin(RegExp(r'(?:^\w|[A-Z]|\b\w|\s+)'),
+      onMatch: (m) => m[0]!.toUpperCase())
+      .replaceAll(' ', '');
+  return '${pascalString}Intl';
+}
+
+String toVariableName(String str) {
```
greglittlefield-wf, line 190: Having doc comments on this and `toNestedName` would be helpful.

lib/src/intl_suggestors/intl_configs_migrator.dart
```diff
@@ -0,0 +1,35 @@
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:codemod/codemod.dart';
+import 'package:file/file.dart';
+import 'package:over_react_codemod/src/intl_suggestors/utils.dart';
+
+const varsToCheck = ['displayName', 'name', 'title'];
+
+class ConfigsMigrator extends RecursiveAstVisitor with AstVisitingSuggestor {
```
greglittlefield-wf, line 9: It looks like this suggestor will operate on any class with these variables, not just config classes.

lib/src/intl_suggestors/constants.dart
```diff
@@ -0,0 +1,17 @@
+const propsToCheck = [
```
greglittlefield-wf, line 1: Is there a reason only certain props are being checked, as opposed to excluding props we don't want to check? Seems like there could be some false negatives.

greglittlefield-wf (approved):
LGTM. Spotted a few things but those can be addressed in follow-up PRs. +1

I'd like to get review from someone else on the internationalization effort before merging, though.

lib/src/executables/intl_message_migration.dart
matthewward-wf:
```Dart
import "package:intl/intl.dart";

//ignore: avoid_classes_with_only_static_members
class CLASSNAME {
```

matthewward-wf: I was doing this via a "wasModified" flag to start, but that wasn't idempotent as a subsequent run would delete the file.

greglittlefield-wf: This seems a bit brittle. What about storing that code as a variable and checking whether the file contains only that code?

lib/src/intl_suggestors/intl_configs_migrator.dart
matthewward-wf: I do not understand why.   

Shouldn't this limit its scope?
```Dart
 @override
  bool shouldSkip(FileContext context) =>
      !context.path.contains('_config.dart');
```

lib/src/intl_suggestors/constants.dart
greglittlefield-wf: That makes sense. That reasoning might be worth capturing as a code comment somewhere.

Though, I could still see blocking props like that (just blocking `size`, `color`, `variant` might get you pretty far). Or maybe it's not so nice in practice 🤷, just thought I'd call it out.

lib/src/intl_suggestors/intl_configs_migrator.dart
greglittlefield-wf: Oh I'm sorry, you're totally right. Never mind!

lib/src/intl_suggestors/utils.dart
matthewward-wf: Like really hard

lib/src/intl_suggestors/constants.dart
matthewward-wf: Yeah.  I think in either case it is a guessing game.  Guess to include user facing string, or guess to exclude non user facing strings.

matthewward-wf:
Ticket for the follow up work: https://jira.atl.workiva.net/browse/INTL-4

lib/src/intl_suggestors/intl_configs_migrator.dart
```diff
@@ -0,0 +1,35 @@
...
+  final String _className;
+  final File _outputFile;
+
+  ConfigsMigrator(this._className, this._outputFile);
+
+  @override
+  void visitMethodDeclaration(MethodDeclaration node) {
+    super.visitMethodDeclaration(node);
+    if (varsToCheck.contains(node.name.name)) {
+      if (node.body is ExpressionFunctionBody) {
```
nicbiggs-wk, line 19: Does Dart do short-circuiting in boolean expressions? If so you could probably put these into one if statement since it doesn't look like anything is an exception to it.

lib/src/executables/intl_message_migration.dart
```diff
@@ -0,0 +1,315 @@
...
+  //
+  // An alternative would be to use `--` and `arguments.rest` to pass along codemod
+  // args, but that's not as convenient to the user and makes showing help a bit more complicated.
+  final codemodArgs = _allCodemodFlags
+      .where((name) => parsedArgs[name] as bool)
+      .map((name) => '--$name')
+      .toList();
+
+  // codemod sets up a global logging handler that forwards to the console, and
+  // we want that set up before we do other non-codemodd things that might log.
```
robbecker-wf, line 109:
```suggestion
  // we want that set up before we do other non-codemod things that might log.
```

lib/src/executables/intl_message_migration.dart
```diff
@@ -0,0 +1,315 @@
...
+  Map<String, String> packageNameLookup = Map.fromIterable(
+    pubspecYamlPaths(),
+    key: (path) => findPackageRootFor(path).split('/').last,
+    value: (path) {
+      final pubspec = fs.file(path);
+      List<String> pubspecLines = pubspec.readAsLinesSync();
+      String nameLine =
+          pubspecLines.firstWhere((line) => line.startsWith('name'));
+      var nameLineParts = nameLine.split(':');
+      return nameLineParts[1].trim();
```
robbecker-wf, line 170: manually reading as lines and extracting the name could be replaced by using the yaml package and parsing using a library like pubspec, pubspec_parse, etc

lib/src/intl_suggestors/intl_configs_migrator.dart
matthewward-wf: I just kept it simple.  I didn't want to end up with
```Dart
if (varsToCheck.contains(node.name.name) && node.body is ExpressionFunctionBody && (body as ExpressionFunctionBody).expression is SimpleStringLiteral) {
```

lib/src/executables/intl_message_migration.dart
matthewward-wf: Yeah, but is it worth pulling in additional deps for 4 lines of code?

robbecker-wf (approved):
+1 .. some optional comments .. could be included in the planned followup.

matthewward-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

lib/src/intl_suggestors/utils.dart
greglittlefield-wf: Oh, I meant in this case that `node.elements.first` and `node.elements.last` have empty `.content`. You're right that the `toString()` representations are "'" though.



For example, for the string interpolation `'$bar'`
```
.elements[0]:
  runtimeType: InterpolationStringImpl
  "toString()":  "'"
  "InterpolationString.value": ""

.elements[1]:
  runtimeType: InterpolationExpressionImpl
  "toString()":  "$bar"

.elements[2]:
  runtimeType: InterpolationStringImpl
  "toString()":  "'"
  "InterpolationString.value": ""
```

<details>
<summary>(source code for that ^)</summary>

```dart
import 'package:analyzer/dart/analysis/utilities.dart';
import 'package:analyzer/dart/ast/visitor.dart';

void main() {
  final unit = parseString(content: r'''
    foo() => '$bar';
  ''').unit;

  unit.accept(MyVisitor());
}

class MyVisitor extends RecursiveAstVisitor<void> {
  @override
  void visitStringInterpolation(StringInterpolation node) {
    for (var i = 0; i < node.elements.length; i++) {
      final element = node.elements[i];
      print('.elements[$i]:');
      print('  runtimeType: ${element.runtimeType}');
      print('  "toString()":  "$element"');
      if (element is InterpolationString) {
        print('  "InterpolationString.value": "${element.value}"');
      }
      print('');
    }
  }
}
```

</details>

So, I was thinking this check would be more precise if it were:
```dart
  if (node.elements.length == 3) {
    final first = node.elements.first;
    final last = node.elements.last;
    if (first is InterpolationString && last is InterpolationString && 
        first.value.isEmpty && last.value.isEmpty) {
      ...
    }
  }
```

PR #160, opened 2022-07-12
INTL-78: Exclude some bad cases from intl_message_migration

alanknight-wk:
## Motivation
  
Among the cases we've run into in usage:
 - SelectOption.value should not be translated, although value is usually a translatable attribute
 - FormComponentDisplayPropsMixin.label is sometimes translated, unless hideLabel is true, in which case it isn't, and seems to be used as a program idenitifer.
 - If a string is a URL, even if it's in a translatable attribute and will be presented to the user, we shouldn't try to translate it.

## Changes
  
For the label case we translate it, but add a FIXME for a manual check. For the other two cases we detect them and do not translate.
There are a lot of format changes in this, I don't know why.

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist


aviary-wf:
## Security Insights

## Action Items
---

lib/src/intl_suggestors/utils.dart
```diff
@@ -277,16 +284,28 @@ String? getTestId(String? testId, AstNode node) {
   }
 }
 
-bool excludeUnlikelyExpressions<E extends Expression>(PropAssignment prop,
-    String propKey) {
+/// Exclude cases we know shouldn't be localized, even though that attribute
```
alanknight-wk, line 287: This is a real change.

lib/src/intl_suggestors/utils.dart
```diff
@@ -3,6 +3,7 @@ import 'package:analyzer/dart/ast/token.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 import 'package:file/file.dart';
+import 'package:over_react_codemod/src/util.dart';
```
alanknight-wk, line 6: I've flagged the real changes in here, to distinguish from the formatting ones.

lib/src/intl_suggestors/utils.dart
```diff
@@ -23,21 +24,27 @@ bool isValidStringLiteralNode(AstNode node) {
   if (node.value.isEmpty) return false;
   if (double.tryParse(node.value) != null) return false;
   if (quotedCamelCase(node.value)) return false;
-  if (node.value
-      .trim()
-      .length == 1) return false;
+  if (node.value.trim().length == 1) return false;
```
alanknight-wk, line 27: This is real - exclude URLS.

lib/src/intl_suggestors/utils.dart
```diff
@@ -23,21 +24,27 @@ bool isValidStringLiteralNode(AstNode node) {
...
+  if (mightBeAUrl != null &&
+      ['http', 'https', 'wurl', 'mailTo'].contains(mightBeAUrl.scheme)) {
+    return false;
+  }
   return true;
 }
 
 bool isValidStringInterpolationProp(PropAssignment prop) {
   if (!propsToCheck.contains(prop.name.name)) return false;
+  if (excludeKnownBadCases(prop, prop.name.name)) return false;
```
alanknight-wk, line 39: This is real.

lib/src/intl_suggestors/utils.dart
```diff
@@ -23,21 +24,27 @@ bool isValidStringLiteralNode(AstNode node) {
...
   if (excludeUnlikelyExpressions(prop, prop.name.name)) return false;
   if (!isValidStringInterpolationNode(prop.rightHandSide)) return false;
   return true;
 }
 
 bool isValidStringLiteralProp(PropAssignment prop) {
   if (!propsToCheck.contains(prop.name.name)) return false;
+  if (excludeKnownBadCases(prop, prop.name.name)) return false;
```
alanknight-wk, line 47: This is real.

lib/src/util.dart
```diff
@@ -435,6 +435,29 @@ extension TypeOrBound on DartType {
   }
 }
 
+extension IsA on DartType {
```
alanknight-wk, line 438: Not sure this is the best way to check this, but it seems to work, and be useful. I even wrote a test for it, since it's going into the more general codemod infrastructure.

test/util_test.dart
```diff
@@ -13,22 +13,48 @@
...
   group('Utils', () {
+    group('isA', () {
+      late ResolvedUnitResult result;
+      CompilationUnit astNode;
+
+      setUp(() async {
+        // final sourceFile = SourceFile.fromString(content)
+        // astNode = parseString(content: content).unit;
+        var path = p.join(Directory.current.absolute.path, 'test',
+            'util_test_things_to_parse.dart');
```
alanknight-wk, line 41: It's hard to set up a resolved AST on a string in memory, so I made a separate file.

lib/src/util.dart
greglittlefield-wf: We have this existing extension which does something very similar: https://github.com/Workiva/over_react_codemod/blob/master/lib/src/util/element_type_helpers.dart#L26

so it might make sense to consolidate them.

test/util_test.dart
greglittlefield-wf: FYI there's `SharedAnalysisContext` which makes it a bit easier. 
Example: 
- Setup: https://github.com/Workiva/over_react_codemod/blob/5e30958979a650d0850fadabb068f6ea17d542a7/test/util/element_type_helpers_test.dart#L57-L67
- Usage: https://github.com/Workiva/over_react_codemod/blob/5e30958979a650d0850fadabb068f6ea17d542a7/test/util/element_type_helpers_test.dart#L87-L106

greglittlefield-wf (commented):
Added some comments about potentially reusing existing code, otherwise looks good!

lib/src/util.dart
alanknight-wk: Oh, yes. I'd seen that, but didn't realize that it actually did what I wanted. Switched to using it, and removed the extension. And now I don't need the test.

greglittlefield-wf (approved):
Utility-related changes and general approach looks good to me! +1

I don't feel comfortable approving the behavioral changes and handling of bad cases though, so I think review from the INTL team is needed before merge.

lib/src/intl_suggestors/intl_migrator.dart
```diff
@@ -12,7 +13,7 @@ class IntlMigrator extends ComponentUsageMigrator {
   IntlMigrator(this._className, this._outputFile);
 
   @override
-  String get fixmePrefix => 'FIXME(intl_migration)';
+  String get fixmePrefix => 'FIXME - INTL ';
```
robbecker-wf, line 16: Thanks, this is the preferred fixme comment format.

lib/src/intl_suggestors/intl_migrator.dart
```diff
@@ -40,7 +41,8 @@ class IntlMigrator extends ComponentUsageMigrator {
     super.migrateUsage(usage);
     final namePrefix =
         usage.node.thisOrAncestorOfType<ClassDeclaration>()?.name.name ??
-                usage.node.thisOrAncestorOfType<VariableDeclaration>()?.name.name ?? 'null';
+            usage.node.thisOrAncestorOfType<VariableDeclaration>()?.name.name ??
+            'null';
```
robbecker-wf, line 45: note: this appears to be just a formatting change

alanknight-wk: Yes, my IDE is set to auto-format, and it seemed to change a lot of stuff. Not clear why, this repo isn't set up for dart_dev, so it's just doing a generic dartfmt

robbecker-wf (approved):
Nice, after using Greg's suggestions this looks good. No further questions.

alanknight-wk:
QA +1
Tested manually, because we don't have a lot of automation on the intl_message_migration.

alanknight-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #191, opened 2022-10-13
INTL-176: Ignore statement or file

jaybillmccarthy-wk:
## Motivation
We need a way to selectively prevent the internationalization codemod from running on specific statements or files

## Changes
Updated codemod to look for specific comments and ignore either the file or the next statement. The next statement logic will ignore a statement across multiple lines.

#### Release Notes
* Add logic that looks for `// IGNORE STATEMENT - INTL` and does no internationalization processing on the next statement
* Add logic that looks for `// IGNORE FILE - INTL` anywhere in a file and does no internationalization processing on the entire file 

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/intl_suggestors/intl_migrator.dart
```diff
@@ -81,6 +81,9 @@ class IntlMigrator extends ComponentUsageMigrator {
 
   IntlMigrator(this._className, this._messages);
 
+  String get ignoreStatement => 'IGNORE STATEMENT - INTL';
```
robbecker-wf, line 84:
```suggestion
  String get ignoreStatement => 'ignore: intl';
```

lib/src/intl_suggestors/intl_migrator.dart
```diff
@@ -81,6 +81,9 @@ class IntlMigrator extends ComponentUsageMigrator {
...
+  String get ignoreFile => 'IGNORE FILE - INTL';
```
robbecker-wf, line 85:
```suggestion
  String get ignoreFile => 'ignore_for_file: intl';
```

robbecker-wf:
I gave it some testing locally and from what I tried I can get it to ignore a whole file but nothing I tried would let me ignore a specific string.

lib/src/intl_suggestors/intl_migrator.dart
jaybillmccarthy-wk: I can change this, but I was trying to follow the pattern set by `// FIXME - INTL`. Should I still change it?

robbecker-wf: Ignoring other things in dart follows this pattern (lints for example) so I think it would be the default format that devs would go to. I like the `// ignore:` prefix but could be convinced to provide a more specific word than just `intl` .... maybe `intl_codemod` or something.

lib/src/intl_suggestors/intl_migrator.dart
alanknight-wk: I'm not sure we should use exactly the same comment as the linter. That might confuse people who would think it's a lint rule, not a directive.

It looks like there are `orcm_ignore` and `orcm_ignore_for_file` comments in the codemod already (see README.md, lib/src/util/ignore_info.dart) but it's not a perfect fit. But maybe we should have a parallel intl_ignore:, or intl_codemod_ignore:?

lib/src/intl_suggestors/intl_migrator.dart
```diff
@@ -97,14 +100,18 @@ class IntlMigrator extends ComponentUsageMigrator {
...
-  bool shouldMigrateUsage(FluentComponentUsage usage) =>
-      // TODO: Handle adjacent strings with an interpolation.
-      usage.cascadedProps.any((prop) =>
-          isValidStringLiteralProp(prop) ||
-          isValidStringInterpolationProp(prop)) ||
-      usage.children.any((child) =>
-          isValidStringLiteralNode(child.node) ||
-          isValidStringInterpolationNode(child.node));
+  bool shouldMigrateUsage(FluentComponentUsage usage) {
+    // TODO: Handle adjacent strings with an interpolation.
```
alanknight-wk, line 104: I think this TODO is obsolete.

lib/src/intl_suggestors/intl_migrator.dart
```diff
@@ -163,6 +170,20 @@ class IntlMigrator extends ComponentUsageMigrator {
     // }
   }
 
+  bool isLineIgnored(node) {
+    final ignoreComment = node.findPrevious(node.beginToken).precedingComments;
```
alanknight-wk, line 174: nit: 
```
final ignoreComment = ....precedingComments ?? '';
return ignoreComment.value().contains(ignoreStatement);
```

lib/src/intl_suggestors/intl_migrator.dart
```diff
@@ -163,6 +170,20 @@ class IntlMigrator extends ComponentUsageMigrator {
...
+    if (ignoreComment != null &&
+        ignoreComment.value().contains(ignoreStatement)) {
+      return true;
+    }
+    return false;
+  }
+
+  bool isFileIgnored(fileContents) {
```
alanknight-wk, line 182: Not to prematurely optimize, but maybe we could pass in the file path and then remember the result for particular files.

lib/src/intl_suggestors/intl_migrator.dart
```diff
@@ -163,6 +170,20 @@ class IntlMigrator extends ComponentUsageMigrator {
...
+    final ignored = fileContents.contains(this.ignoreFile);
```
alanknight-wk, line 183: unnecessary `this.`, as above.

lib/src/intl_suggestors/intl_migrator.dart
```diff
@@ -97,14 +100,18 @@ class IntlMigrator extends ComponentUsageMigrator {
...
-      usage.cascadedProps.any((prop) =>
-          isValidStringLiteralProp(prop) ||
-          isValidStringInterpolationProp(prop)) ||
-      usage.children.any((child) =>
-          isValidStringLiteralNode(child.node) ||
-          isValidStringInterpolationNode(child.node));
+  bool shouldMigrateUsage(FluentComponentUsage usage) {
+    // TODO: Handle adjacent strings with an interpolation.
+    if (isLineIgnored(usage.node)) return false;
+    if (isFileIgnored(this.context.sourceText)) return false;
```
alanknight-wk, line 106: You don't need the `this.`, it's implicit in Dart.

lib/src/intl_suggestors/intl_migrator.dart
```diff
@@ -163,6 +170,20 @@ class IntlMigrator extends ComponentUsageMigrator {
...
+    if (ignoreComment != null &&
+        ignoreComment.value().contains(ignoreStatement)) {
+      return true;
+    }
+    return false;
+  }
+
+  bool isFileIgnored(fileContents) {
+    final ignored = fileContents.contains(this.ignoreFile);
+    return ignored;
```
alanknight-wk, line 184: Also, I think we'll want to move these outside the specific migrator. There are 4 we use -ConfigsMigrator, IntlMigrator, ConstantStringsMigrator (and intlImporter, but I don't think this applies to that one). So we'll probably want as much as possible to be generic.

test/intl_suggestors/intl_migrator_test.dart
```diff
@@ -1299,5 +1299,132 @@ void main() {
         expect(messages.messageContents(), expectedFileContent);
       });
     });
+
+    group('Ignore', () {
+      test('Ignore statement with ignore comment', () async {
```
alanknight-wk, line 1304: As in the other test cases below, triple-quoted strings are really useful for this kind of thing.

Also, in Dart, you can use either single or double quoted strings for the outside (and single triple-quotes or double-triple quotes). Normal convention is to use single-quotes, but when the thing contained is Dart code, that may have its own quoted strings, I prefer to use double quotes for the outer, so the inner part looks more like normal Dart code.

Finally, note that you can have raw strings, so if you don't have any interpolations, you can write
   `r"=> Intl.message('blah $stuff'...);"`
and you don't have to escape the interpolations. It also ignores backslashes, so you can more easily have literal backslashes as well.

test/intl_suggestors/intl_migrator_test.dart
```diff
@@ -1299,5 +1299,132 @@ void main() {
...
+            ');\n'
+            '';
+
+        await testSuggestor(
+          input: source,
+          expectedOutput: output,
+        );
+      });
+
+      test('Ignore line in chained objects', () async {
```
alanknight-wk, line 1362: nit: 'Ignore comment ...

test/intl_suggestors/intl_migrator_test.dart
```diff
@@ -1299,5 +1299,132 @@ void main() {
...
+        }
+
+        class Baz {
+          Bar newBar() {
+            return Bar();
+          }
+        }
+
+        class Biz {
+          static const message = "Message";
```
alanknight-wk, line 1386: As per chat, this test's setup only sets up the IntlMigrator, which isn't what changes constants.

lib/src/intl_suggestors/intl_migrator.dart
jaybillmccarthy-wk: Oh, wow, I didn't know you could do that. I changed this whole thing, so this comment doesn't apply, but I will totally do that in this situation in the future.

lib/src/intl_suggestors/intl_migrator.dart
jaybillmccarthy-wk: changed these to:  `//ignore_statement: intl_message_migration` and `//ignore_file: intl_message_migration` respectively.

lib/src/intl_suggestors/intl_migrator.dart
jaybillmccarthy-wk: See below

lib/src/intl_suggestors/intl_migrator.dart
alanknight-wk: Well this is interesting. I went to look at how one might do this and came across lib/src/util/ignore_info.dart, which is doing basically the same thing, including finding the preceding comment, for the orcm_ignore and orcm_ignore_for_file comments.

And it's using, like, *all* the crazy Dart tricks at once. A static extension on CompilationUnit, a sync* method for finding the comments, and an Expando for remembering the ignore comment status. But I'm not sure we want to adapt or clone that code directly, but it might worth looking at for anything missed.

That part doesn't seem to do anything about skipping the entire file without visiting individual uses, but it does seem like there's an inherited 
  bool shouldSkip(FileContext context) => false;
which could possibly be overridden to check for the ignore_for_file comment.

.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
...
+            "request": "launch",
+            "type": "dart",
+            "program": "bin/intl_message_migration.dart",
+            "args": [
+                "--yes-to-all",
+                "../wdesk_sdk/lib/src/app_infrastructure/error_dev_tool_module/exception_modal_component.dart"
+            ]
+        }
+    ]
+}
```
robbecker-wf, line 18: can remove this file

lib/src/intl_suggestors/intl_migrator.dart
```diff
@@ -163,6 +169,46 @@ class IntlMigrator extends ComponentUsageMigrator {
...
+
+    if (node.precedingComments != null &&
+        node.precedingComments.value().contains(ignoreStatement)) {
+      return true;
+    } else {
+      return isIgnoreCommentBeforePreviousTerminator(node.previous, limit - 1);
+    }
+  }
+
+  bool isLineIgnored(node) {
```
alanknight-wk, line 196: It seems superficially like this should work for comments within the statement, as with the below example I tried in publishing_client. But it doesn't seem to. If it's hard, we might reasonably leave it for a follow-on PR.
```
 return (Dom.div())(
      // ignore_statement: intl_message_migration
      (Dom.p())(
        'This template will appear in a new ',
        (Dom.span()..className = 'text-bold')(_destinationFolderName()),
        ' folder, which only you can use unless you add collaborators.',
      ),
      (Dom.p())(
        'The template will install in the background while you work; we\'ll let you know when it\'s ready!',
      ),
    );
```

.vscode/launch.json
jaybillmccarthy-wk: removed

lib/src/intl_suggestors/intl_migrator.dart
jaybillmccarthy-wk: @robbecker-wf and I just experimented with that exact case, actually. It doesn't work, but I agree that it absolutely should. We decided that we'll merge this with a known issue and then add a new ticket to address the shortcomings of this version.

robbecker-wf (approved):
Some cleanups and improvements can be done in a follow up.

robbecker-wf:
QA+1 CI Passes and tested locally. Works well enough with some known caveats of comment placement for statements.

robbecker-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #193, opened 2022-11-04
INTL-113 : ignore initial variable names in creating selectors

chetansinghchoudhary-wk:
## Purpose
  The codemod uses the first few words of the literal string to create a selector. But if the first few words are an interpolated variable, then for
'$numHiddenPills additional values not shown'
we get
GraphUiIntl.GraphMultiEdgeInputPrimitiveComponent_numHiddenPills

It would be better if it ignored the variables for this.
## Solution
Add the condition which check if string starts with interpolated variable so it will remove the interpolated variable from start 

## Semantic Versioning (check one)

- [ ] The following were changed in a non-backward compatible way and requires a major version bump:
  - *[link to the breaking change in the diff]*
- [ ] Something public was added or changed in backward compatible way, this requires a minor version bump
- [ ] No public changes nor new features (backwards-compatible refactor or bug fix), so this can be included in a patch release

## How to QA

- [ ] CI tests exercise all new functionality

### DIFF
--- a/lib/src/intl_suggestors/intl_message_syntax.dart
+++ b/lib/src/intl_suggestors/intl_message_syntax.dart
@@ -146,9 +146,10 @@ class MessageSyntax {
     // Needing condition initialName.endsWith('_intlFunction') bcz when any node has addTestId we need to give that
     // priority, and will not override the node name.
     if (body is StringInterpolation && (initialName == null || initialName.endsWith('_intlFunction'))) {
-      var strings = body.elements.where((each) => each is InterpolationString).map((each) => each.toSource()).toList();
+
+      var strings = body.elements.where((each) => each is InterpolationString).map((each) => (each as InterpolationString).value).toList();
       var data=strings.join(' ').trim();
-      if(data.isNotEmpty){
+      if(data.isNotEmpty && !data.contains("'")){
         String name = toVariableName(data);
         if(name.trim().isNotEmpty){
           var functionName = owner.nameForString(name, messageText,
--- a/lib/src/intl_suggestors/intl_message_syntax.dart
+++ b/lib/src/intl_suggestors/intl_message_syntax.dart
@@ -52,18 +52,10 @@ class MessageSyntax {
   /// A template to build a function name for Intl message
   /// ex: Foo_bar
   String functionNameFor(
-    StringInterpolation node,
-    String namePrefix,
-  ) {
-    // checks for newline in string
-    var ignoreNewLineString = node.toSource();
-    if (ignoreNewLineString.contains(r'\n')) {
-      return '${namePrefix}_intlFunction';
-    }
-    return '${getTestId(null, node) ?? '${namePrefix}_intlFunction'}';
-  }
-
-
+      StringInterpolation node,
+      String namePrefix,
+      ) =>
+      '${getTestId(null, node) ?? '${namePrefix}_intlFunction'}';
 
   /// Returns Intl.message with interpolation
   /// ex: static String Foo_bar(String baz) => Intl.message(
@@ -147,20 +139,22 @@ class MessageSyntax {
     return '(${args.join(', ')})';
   }
 
-
   String nameForNode(StringLiteral body,
       {String? initialName, bool startAtZero = false}) {
     var messageText = body.toSource();
-    var functionName;
+
+    //Strings from constant fields will not pass below condition so initialName will be unchanged.
     if (body is StringInterpolation) {
       var strings = body.elements.where((each) => each is InterpolationString).map((each) => each.toSource()).toList();
       var data=strings.join(' ').replaceAll("'", '').trim();
-      if(data.isNotEmpty && data.contains(' ')){
-        functionName=toVariableName(data);
-        initialName=functionName;
+      if(data.isNotEmpty){
+        var name = toVariableName(data);
+        var functionName = owner.nameForString(name, messageText,
+            startAtZero: startAtZero);
+        return functionName;
       }
     }
-      functionName = toVariableName(messageText);
+    var functionName = toVariableName(messageText);
     functionName = owner.nameForString(initialName ?? functionName, messageText,
         startAtZero: startAtZero);
     return functionName;
--- a/lib/src/intl_suggestors/intl_message_syntax.dart
+++ b/lib/src/intl_suggestors/intl_message_syntax.dart
@@ -1,4 +1,5 @@
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:intl/intl.dart';
 import 'package:over_react_codemod/src/intl_suggestors/intl_messages.dart';
 import 'package:over_react_codemod/src/intl_suggestors/utils.dart';
 
@@ -40,15 +41,13 @@ class MessageSyntax {
     String namespace,
     String namePrefix,
   ) {
-    final baseName = functionNameFor(node, namePrefix);
-    // TODO: This is very messy to accomodate names starting at intlFunction0
-    // for backward compatibilitys.
-    final functionName = nameForNode(node,
-        initialName: baseName, startAtZero: baseName.endsWith('_intlFunction'));
+    // Generating function name directly by String content rather then testId or appending _intlFunction
+    final functionName = nameForNode(node);
     final functionArgs = intlFunctionArguments(node);
     return '$namespace.$functionName$functionArgs';
   }
 
+  //TODO: We can remove this method definition `getTestId` as we are using string content now to generate intl function name.
   /// A template to build a function name for Intl message
   /// ex: Foo_bar
   String functionNameFor(
@@ -65,11 +64,8 @@ class MessageSyntax {
   ///                                           );
   String functionDefinition(
       StringInterpolation node, String namespace, String namePrefix) {
-    final baseName = functionNameFor(node, namePrefix);
-    // TODO: This is very messy to accomodate function names starting at
-    // intlFunction0 for backward-compatibility.
-    final functionName = nameForNode(node,
-        initialName: baseName, startAtZero: baseName.endsWith('_intlFunction'));
+    // Generating function name directly by String content rather then testId or appending _intlFunction
+    final functionName = nameForNode(node);
     final functionParams = intlFunctionParameters(node);
     final parameterizedMessage = intlParameterizedMessage(node);
     final messageArgs = intlMessageArgs(node);
@@ -139,31 +135,24 @@ class MessageSyntax {
     return '(${args.join(', ')})';
   }
 
+  /// For a string interpolation, get all the non-interpolated parts in a string,
+  /// separated by spaces.
+  String textFromInterpolation(StringInterpolation body) => body.elements
+      .whereType<InterpolationString>()
+      .map((each) => each.value)
+      .join(' ')
+      .trim();
+
   String nameForNode(StringLiteral body,
-      {String? initialName, bool startAtZero = false}) {
-    var messageText = body.toSource();
-    //Strings from constant fields will not pass below condition so initialName will be unchanged.
-    // Needing condition initialName.endsWith('_intlFunction') bcz when any node has addTestId we need to give that
-    // priority, and will not override the node name.
-    if (body is StringInterpolation &&
-        (initialName == null || initialName.endsWith('_intlFunction'))) {
-      var strings = body.elements
-          .where((each) => each is InterpolationString)
-          .map((each) => (each as InterpolationString).value)
-          .toList();
-      var data = strings.join(' ').trim();
-      if (data.isNotEmpty) {
-        String name = toVariableName(data);
-        if (name.trim().isNotEmpty) {
-          var functionName =
-              owner.nameForString(name, messageText, startAtZero: false);
-          return functionName;
-        }
-      }
-    }
-    var functionName = toVariableName(messageText);
-    functionName = owner.nameForString(initialName ?? functionName, messageText,
-        startAtZero: startAtZero);
-    return functionName;
+      {String? initialName=null, bool startAtZero = false}) {
+    var messageText = body is StringInterpolation
+        ? textFromInterpolation(body)
+        : body.toSource();
+    // The case where there is no text after this should be checked in
+    // isValidStringInterpolationNode and so it should never happen here.
+    return owner.nameForString(initialName ?? toVariableName(messageText), messageText,
+         startAtZero: startAtZero);
+
   }
+
 }
--- a/lib/src/intl_suggestors/intl_message_syntax.dart
+++ b/lib/src/intl_suggestors/intl_message_syntax.dart
@@ -135,13 +135,7 @@ class MessageSyntax {
     return '(${args.join(', ')})';
   }
 
-  /// For a string interpolation, get all the non-interpolated parts in a string,
-  /// separated by spaces.
-  String textFromInterpolation(StringInterpolation body) => body.elements
-      .whereType<InterpolationString>()
-      .map((each) => each.value)
-      .join(' ')
-      .trim();
+
 
   String nameForNode(StringLiteral body,
       {String? initialName=null, bool startAtZero = false}) {
@@ -150,8 +144,10 @@ class MessageSyntax {
         : body.toSource();
     // The case where there is no text after this should be checked in
     // isValidStringInterpolationNode and so it should never happen here.
-    return owner.nameForString(initialName ?? toVariableName(messageText), messageText,
-         startAtZero: startAtZero);
+return owner.nameForString(
+          initialName ?? toVariableName(messageText), messageText,
+          startAtZero: startAtZero);
+
 
   }
 
--- a/lib/src/intl_suggestors/intl_message_syntax.dart
+++ b/lib/src/intl_suggestors/intl_message_syntax.dart
@@ -142,16 +142,19 @@ class MessageSyntax {
   String nameForNode(StringLiteral body,
       {String? initialName, bool startAtZero = false}) {
     var messageText = body.toSource();
-
     //Strings from constant fields will not pass below condition so initialName will be unchanged.
-    if (body is StringInterpolation) {
+    // Needing condition initialName.endsWith('_intlFunction') bcz when any node has addTestId we need to give that
+    // priority, and will not override the node name.
+    if (body is StringInterpolation && (initialName == null || initialName.endsWith('_intlFunction'))) {
       var strings = body.elements.where((each) => each is InterpolationString).map((each) => each.toSource()).toList();
-      var data=strings.join(' ').replaceAll("'", '').trim();
+      var data=strings.join(' ').trim();
       if(data.isNotEmpty){
-        var name = toVariableName(data);
-        var functionName = owner.nameForString(name, messageText,
-            startAtZero: startAtZero);
-        return functionName;
+        String name = toVariableName(data);
+        if(name.trim().isNotEmpty){
+          var functionName = owner.nameForString(name, messageText,
+              startAtZero: false);
+          return functionName;
+        }
       }
     }
     var functionName = toVariableName(messageText);
--- a/lib/src/intl_suggestors/intl_message_syntax.dart
+++ b/lib/src/intl_suggestors/intl_message_syntax.dart
@@ -41,7 +41,6 @@ class MessageSyntax {
     String namePrefix,
   ) {
     final baseName = functionNameFor(node, namePrefix);
-    print("Name prefix $namePrefix");
     // TODO: This is very messy to accomodate names starting at intlFunction0
     // for backward compatibilitys.
     final functionName = nameForNode(node,
@@ -146,7 +145,6 @@ class MessageSyntax {
     //Strings from constant fields will not pass below condition so initialName will be unchanged.
     // Needing condition initialName.endsWith('_intlFunction') bcz when any node has addTestId we need to give that
     // priority, and will not override the node name.
-    //(initialName == null || initialName.endsWith('_intlFunction'))
     if (body is StringInterpolation &&
         (initialName == null || initialName.endsWith('_intlFunction'))) {
       var strings = body.elements
--- a/lib/src/intl_suggestors/intl_message_syntax.dart
+++ b/lib/src/intl_suggestors/intl_message_syntax.dart
@@ -55,12 +55,15 @@ class MessageSyntax {
     StringInterpolation node,
     String namePrefix,
   ) {
+    // checks for newline in string
     var ignoreNewLineString = node.toSource();
     if (ignoreNewLineString.contains(r'\n')) {
       return '${namePrefix}_intlFunction';
     }
     return '${getTestId(null, node) ?? '${namePrefix}_intlFunction'}';
   }
+
+
 
   /// Returns Intl.message with interpolation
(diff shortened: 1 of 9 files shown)


aviary-wf:
## Security Insights

## Action Items
---

lib/src/intl_suggestors/utils.dart
```diff
@@ -142,8 +142,16 @@ String toVariableName(String str) {
   if (str.startsWith("'") && str.endsWith("'")) {
     str = str.substring(1, str.length - 1);
   }
+  // This will ignore the first few interpolated variable
```
alanknight-wk, line 145: Rather than doing this here, where it's already been turned into a string, so there's less information, consider doing it where it's called, when we actually have the StringInterpolation AstNode. I think `nameForNode` in intl_message_syntax.dart looks promising. Then you can ask a StringInterpolation for e.g. elements and discard the interpolations.

lib/src/intl_suggestors/intl_message_syntax.dart
```diff
@@ -139,9 +144,30 @@ class MessageSyntax {
...
+    return messageText;
+  }
+
   String nameForNode(StringLiteral body,
       {String? initialName, bool startAtZero = false}) {
     var messageText = body.toSource();
+    var removingInterpolatedVariable = ignoreInterpolatedVariable(messageText);
+    if (removingInterpolatedVariable.isNotEmpty &&
+        removingInterpolatedVariable.contains(' ')) {
+      initialName = toVariableName(removingInterpolatedVariable);
```
alanknight-wk, line 169: This is the reason the constant tests are failing. You're overwriting the parameter initialName, and in those tests it's being passed in.

lib/src/intl_suggestors/intl_message_syntax.dart
```diff
@@ -54,8 +54,13 @@ class MessageSyntax {
   String functionNameFor(
     StringInterpolation node,
     String namePrefix,
-  ) =>
-      '${getTestId(null, node) ?? '${namePrefix}_intlFunction'}';
+  ) {
+    var ignoreNewLineString = node.toSource();
```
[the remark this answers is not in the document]
chetansinghchoudhary-wk: Actually, here our `single line with an explicit new line` test case is failing [here](https://github.com/Workiva/over_react_codemod/blob/master/test/intl_suggestors/utils_test.dart#L104)  that's why we added a condition if we have a new line inside the string so it will ignore it and create function name with intlFuction0

lib/src/intl_suggestors/intl_message_syntax.dart
```diff
@@ -139,9 +144,30 @@ class MessageSyntax {
...
+          .substring(messageText.indexOf('\$'), messageText.indexOf(" "));
+      messageText = messageText.replaceFirst(toRemoveFromName, "").trim();
+    }
+    return messageText;
+  }
+
   String nameForNode(StringLiteral body,
       {String? initialName, bool startAtZero = false}) {
     var messageText = body.toSource();
+    var removingInterpolatedVariable = ignoreInterpolatedVariable(messageText);
```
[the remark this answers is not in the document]
chetansinghchoudhary-wk: Code updated as per your suggestion

lib/src/intl_suggestors/intl_message_syntax.dart
chetansinghchoudhary-wk: I updated the code based on your suggestion as [above](https://github.com/Workiva/over_react_codemod/pull/193#discussion_r1015730786) and now `constant tests` are running successfully.

lib/src/intl_suggestors/intl_message_syntax.dart
```diff
@@ -139,10 +147,20 @@ class MessageSyntax {
...
   }
 
+
   String nameForNode(StringLiteral body,
       {String? initialName, bool startAtZero = false}) {
     var messageText = body.toSource();
-    var functionName = toVariableName(messageText);
+    var functionName;
+    if (body is StringInterpolation) {
+      var strings = body.elements.where((each) => each is InterpolationString).map((each) => each.toSource()).toList();
```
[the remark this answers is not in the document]
alanknight-wk: No, I'm wrong. InterpolationString is for the strings, InterpolationExpression is for interpolated values. Never mind.

lib/src/intl_suggestors/intl_message_syntax.dart
```diff
@@ -142,6 +142,22 @@ class MessageSyntax {
...
       {String? initialName, bool startAtZero = false}) {
     var messageText = body.toSource();
+    //Strings from constant fields will not pass below condition so initialName will be unchanged.
+    // Needing condition initialName.endsWith('_intlFunction') bcz when any node has addTestId we need to give that
+    // priority, and will not override the node name.
+    if (body is StringInterpolation && (initialName == null || initialName.endsWith('_intlFunction'))) {
+
+      var strings = body.elements.where((each) => each is InterpolationString).map((each) => (each as InterpolationString).value).toList();
+      var data=strings.join(' ').trim();
+      if(data.isNotEmpty && !data.contains("'")){
```
alanknight-wk, line 152: Do we need the check for apostrophe? It's not really part of the ticket request, but it seems to me that in the test examples we get better results if we do the same thing with or without an apostrophe. e.g. bobsLastName rather than intlFunction0. And then I think this function might be able to be simplified.

lib/src/intl_suggestors/intl_message_syntax.dart
```diff
@@ -142,6 +142,22 @@ class MessageSyntax {
   String nameForNode(StringLiteral body,
       {String? initialName, bool startAtZero = false}) {
     var messageText = body.toSource();
+    //Strings from constant fields will not pass below condition so initialName will be unchanged.
+    // Needing condition initialName.endsWith('_intlFunction') bcz when any node has addTestId we need to give that
+    // priority, and will not override the node name.
+    if (body is StringInterpolation && (initialName == null || initialName.endsWith('_intlFunction'))) {
```
alanknight-wk, line 148: This works, but I think we could have better code structure by backing up a little bit. The _intlFunction name gets passed in from functionCall and functionDefinition, where they call functionNameFor ahead of this, and that's hard-coded to return the name with _intlFunction in it, or else use the test Id. 

If we can create a name from the strings parts of the interpolation, then that's better than either the test id or _intlFunction. So maybe we can not pass in the initialName in that case, and only call the functionNameFor in here as a failure case.

lib/src/intl_suggestors/intl_message_syntax.dart
```diff
@@ -142,6 +142,22 @@ class MessageSyntax {
...
+      var data=strings.join(' ').trim();
+      if(data.isNotEmpty && !data.contains("'")){
+        String name = toVariableName(data);
+        if(name.trim().isNotEmpty){
+          var functionName = owner.nameForString(name, messageText,
+              startAtZero: false);
+          return functionName;
+        }
+      }
+    }
```
alanknight-wk, line 160: Also, for failure cases, if 'data' is empty, then we're falling through to the default case for an interpolation. That should probably be more explicit. I think it probably works right now because in those cases initialName is never null, so the functionName from the interpolation is unused.

lib/src/intl_suggestors/intl_message_syntax.dart
chetansinghchoudhary-wk: Yea, I was just preventing this for the same reason. I can update this to get the same value with or without the apostrophe.

chetansinghchoudhary-wk: This has been updated in latest commit.

lib/src/intl_suggestors/intl_message_syntax.dart
chetansinghchoudhary-wk: Some thoughts on the feedback, Please suggest.
I have applied these conditions to cover the following cases :

- If some values/DOM don't have `testID` and have only Variable in interpolation with space like: `'${someVariable} '`, This is rare but still possible. To cover this case we may have the method name as `_intlFunction`.
- In getter calls, we are receiving node as `null` in `nameForNode`.

If we update the `functionNameFor` method and remove the `_intlFunction` name generation, we might have some method names as null.

alanknight-wk: Obviously, if we do this, then we'd also need to change the callers to not pass the initialName. In this situation I don't think we ever need to fall back to intlFunction0 or similar.

lib/src/intl_suggestors/intl_message_syntax.dart
chetansinghchoudhary-wk: Updated this failure case also.

lib/src/intl_suggestors/utils.dart
```diff
@@ -9,13 +9,25 @@ import 'package:over_react_codemod/src/util/element_type_helpers.dart';
...
+}
 bool isValidStringInterpolationNode(AstNode node) {
   if (node is! StringInterpolation) return false;
   //We do not need to localize single values.  This should be handled by the
   // variable being passed in.  IE: ..foo = '$bar' becomes [', '$bar', ']
   if (node.elements.length == 3 &&
       node.elements.first.toString() == node.elements.last.toString())
     return false;
+  var result=textFromInterpolation(node);
+  if(!result.contains(RegExp('[a-zA-Z0-9]')) || result.isEmpty) return false;
```
alanknight-wk, line 30: RegExp's are fairly expensive to create, and we call this a lot, so it's probably better to put this in a top-level variable.

And for clarity I might write this the opposite way around

`return result.isNotEmpty && result.contains(...)`

lib/src/intl_suggestors/utils.dart
```diff
@@ -9,13 +9,25 @@ import 'package:over_react_codemod/src/util/element_type_helpers.dart';
...
+      .trim();
+}
 bool isValidStringInterpolationNode(AstNode node) {
   if (node is! StringInterpolation) return false;
   //We do not need to localize single values.  This should be handled by the
   // variable being passed in.  IE: ..foo = '$bar' becomes [', '$bar', ']
   if (node.elements.length == 3 &&
       node.elements.first.toString() == node.elements.last.toString())
     return false;
+  var result=textFromInterpolation(node);
```
alanknight-wk, line 29: Do we have a test that exercises this?

lib/src/intl_suggestors/intl_message_syntax.dart
```diff
@@ -1,4 +1,5 @@
 import 'package:analyzer/dart/ast/ast.dart';
+import 'package:intl/intl.dart';
```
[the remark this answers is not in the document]
chetansinghchoudhary-wk: Removed unused import from here

lib/src/intl_suggestors/intl_message_syntax.dart
```diff
@@ -40,21 +41,19 @@ class MessageSyntax {
     String namespace,
     String namePrefix,
   ) {
-    final baseName = functionNameFor(node, namePrefix);
-    // TODO: This is very messy to accomodate names starting at intlFunction0
-    // for backward compatibilitys.
-    final functionName = nameForNode(node,
-        initialName: baseName, startAtZero: baseName.endsWith('_intlFunction'));
+    // Generating function name directly by String content rather then testId or appending _intlFunction
```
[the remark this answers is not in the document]
chetansinghchoudhary-wk: Removed comment from here

lib/src/intl_suggestors/intl_message_syntax.dart
```diff
@@ -65,11 +64,8 @@ class MessageSyntax {
   ///                                           );
   String functionDefinition(
       StringInterpolation node, String namespace, String namePrefix) {
-    final baseName = functionNameFor(node, namePrefix);
-    // TODO: This is very messy to accomodate function names starting at
-    // intlFunction0 for backward-compatibility.
-    final functionName = nameForNode(node,
-        initialName: baseName, startAtZero: baseName.endsWith('_intlFunction'));
+    // Generating function name directly by String content rather then testId or appending _intlFunction
```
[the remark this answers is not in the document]
chetansinghchoudhary-wk: Removed comment from here

lib/src/intl_suggestors/utils.dart
chetansinghchoudhary-wk: Code updated as per your suggestion

lib/src/intl_suggestors/intl_message_syntax.dart
```diff
@@ -40,21 +41,19 @@ class MessageSyntax {
...
-    // for backward compatibilitys.
-    final functionName = nameForNode(node,
-        initialName: baseName, startAtZero: baseName.endsWith('_intlFunction'));
+    // Generating function name directly by String content rather then testId or appending _intlFunction
+    final functionName = nameForNode(node);
     final functionArgs = intlFunctionArguments(node);
     return '$namespace.$functionName$functionArgs';
   }
 
+  //TODO: We can remove this method definition `getTestId` as we are using string content now to generate intl function name.
```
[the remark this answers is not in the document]
chetansinghchoudhary-wk: Deleted getTestId method

test/intl_suggestors/intl_migrator_test.dart
```diff
@@ -1236,15 +1236,15 @@ void main() {
                                  
                   return (Dom.p() 
                     ..addTestId('truss.aboutWdeskModal.versionInfo')
-                    ..label = TestClassIntl.versionInfo(props.version))();
+                    ..label = TestClassIntl.version(props.version))();
```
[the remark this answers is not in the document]
chetansinghchoudhary-wk: Removed this test case from here

lib/src/intl_suggestors/utils.dart
chetansinghchoudhary-wk: Yes I am Working on it

lib/src/intl_suggestors/utils.dart
```diff
@@ -8,15 +8,25 @@ import 'package:over_react_codemod/src/util/element_type_helpers.dart';
...
 bool isValidStringInterpolationNode(AstNode node) {
   if (node is! StringInterpolation) return false;
   //We do not need to localize single values.  This should be handled by the
   // variable being passed in.  IE: ..foo = '$bar' becomes [', '$bar', ']
   if (node.elements.length == 3 &&
       node.elements.first.toString() == node.elements.last.toString())
     return false;
-  return true;
+   result=textFromInterpolation(node);
+  return result.isNotEmpty && result.contains(RegExp('[a-zA-Z]'));
```
alanknight-wk, line 29: The thing I was suggesting to pull out into a top-level variable was the RegExp, not the result variable. e.g.

RegExp alphabetMatcher = RegExp('[a-zA-Z]');

lib/src/mui_suggestors/mui_importer.dart
```diff
@@ -16,11 +16,9 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/source/line_info.dart';
 import 'package:codemod/codemod.dart';
 import 'package:collection/collection.dart';
-import 'package:logging/logging.dart';
```
alanknight-wk, line 19: This is now deleting some unused loggers. I think we should not do that for things outside of the intl_suggestors.

test/executables/intl_message_migration_test.dart
```diff
@@ -23,7 +23,6 @@ import 'mui_migration_test.dart' show testCodemod;
 
 // Change this to `true` and all of the functional tests in this file will print
 // the stdout/stderr of the codemod processes.
-final _debug = false;
```
alanknight-wk, line 26: It seems this is unused, but if we're deleting it, we should also delete the comment.

lib/src/intl_suggestors/utils.dart
chetansinghchoudhary-wk: Code updated as per your sugesstion

test/executables/intl_message_migration_test.dart
chetansinghchoudhary-wk: Removed unused comment from here

lib/src/mui_suggestors/mui_importer.dart
chetansinghchoudhary-wk: Reverted code added `import 'package:logging/logging.dart';` here

lib/src/mui_suggestors/unused_wsd_import_remover.dart
```diff
@@ -15,9 +15,7 @@
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/syntactic_entity.dart';
 import 'package:codemod/codemod.dart';
-import 'package:logging/logging.dart';
```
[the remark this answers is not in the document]
chetansinghchoudhary-wk: Code updated as per your suggestion

lib/src/intl_suggestors/utils.dart
```diff
@@ -8,15 +8,25 @@ import 'package:over_react_codemod/src/util/element_type_helpers.dart';
 
 import 'constants.dart';
 
+RegExp alphabetMatcher = RegExp('[a-zA-Z]');
```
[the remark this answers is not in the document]
chetansinghchoudhary-wk: Code updated as per your suggestion

lib/src/intl_suggestors/intl_message_syntax.dart
```diff
@@ -139,12 +123,20 @@ class MessageSyntax {
...
-    var functionName = toVariableName(messageText);
-    functionName = owner.nameForString(initialName ?? functionName, messageText,
-        startAtZero: startAtZero);
-    return functionName;
+    var messageText = body is StringInterpolation
+        ? textFromInterpolation(body)
+        : body.toSource();
+    // The case where there is no text after this should be checked in
+    // isValidStringInterpolationNode and so it should never happen here.
+return owner.nameForString(
```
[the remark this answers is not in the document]
chetansinghchoudhary-wk: Run dart format command inside the `lib/src/intl_suggestors` and `test/intl_suggestors`

test/intl_suggestors/utils_test.dart
```diff
@@ -111,9 +112,8 @@ void main() {
 
       test('multiline', () async {
         final testStr = r"'''${multiline}'''";
-        final expectedResult =
-            "  static String NamePrefix_intlFunction0(String multiline) => Intl.message('''\${multiline}''', args: [multiline], name: 'Namespace_NamePrefix_intlFunction0');";
-        runResults(testStr, true, expectedResult);
+        final expectedResult =r"'''${multiline}'''";
+        expect(testStr, expectedResult);
```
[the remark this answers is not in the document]
chetansinghchoudhary-wk: Done, Code reverted from here

test/intl_suggestors/intl_migrator_test.dart
```diff
@@ -179,6 +179,73 @@ void main() {
         );
       });
 
+// here we can call the ignore intrepolated variable string
```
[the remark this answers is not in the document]
chetansinghchoudhary-wk: Done

test/intl_suggestors/intl_migrator_test.dart
```diff
@@ -767,11 +874,11 @@ void main() {
         );
 
         String expectedFileContent1 =
-            "\n  static String foo(String displayName) => Intl.message('Create one from any \${displayName} by selecting Save As Template', args: [displayName], name: 'TestClassIntl_foo');";
+            "\n  static String createOneFromAnyBy(String displayName) => Intl.message('Create one from any \${displayName} by selecting Save As Template', args: [displayName], name: 'TestClassIntl_createOneFromAnyBy');";
         String expectedFileContent2 =
-            "\n  static String bar(String displayName) => Intl.message('Create one from any \${displayName} by selecting Save As Template', args: [displayName], name: 'TestClassIntl_bar');";
+            "\n  static String createOneFromAnyBy1(String displayName) => Intl.message('Create one from any \${displayName} by selecting Save As Template', args: [displayName], name: 'TestClassIntl_createOneFromAnyBy1');";
```
nicbiggs-wk, line 879: This is more a thought that either comes from misunderstanding or probably would be work for another task, but these two message strings look identical so it feels like there shouldn't be a `createOneFromAnyBy1` method and should instead be just one. I don't think this is blocking or anything just something I noticed.

test/intl_suggestors/utils_test.dart
```diff
@@ -87,33 +86,44 @@ void main() {
...
+      void ignoreSingleLineAndMultiline(
+          String testStr, bool isMultiline, String expectedResult) async {
+        final parsedExpression = await sharedContext.parseExpression(testStr);
+        expect(parsedExpression is StringInterpolation, isTrue);
+        var parsedInterpolation = parsedExpression as StringInterpolation;
+        expect(parsedInterpolation.isMultiline, isMultiline);
+        var data = textFromInterpolation(parsedInterpolation);
+        if (data.isEmpty) {
+          data = expectedResult;
+          expect(testStr, expectedResult);
```
nicbiggs-wk, line 104: Isn't this the same expect as the one below this if statement?

test/intl_suggestors/utils_test.dart
```diff
@@ -87,33 +86,44 @@ void main() {
...
 
+      void ignoreSingleLineAndMultiline(
+          String testStr, bool isMultiline, String expectedResult) async {
+        final parsedExpression = await sharedContext.parseExpression(testStr);
+        expect(parsedExpression is StringInterpolation, isTrue);
+        var parsedInterpolation = parsedExpression as StringInterpolation;
+        expect(parsedInterpolation.isMultiline, isMultiline);
+        var data = textFromInterpolation(parsedInterpolation);
+        if (data.isEmpty) {
+          data = expectedResult;
```
nicbiggs-wk, line 103: Is this data variable used anywhere else in the function? it doesn't seem to need to be reset here.

nicbiggs-wk (dismissed):
A few comments, but I don't think any of them are blocking.

test/intl_suggestors/intl_migrator_test.dart
chetansinghchoudhary-wk: This test case is intentional to show method names for similar strings, In current PR we are not reusing the existing function via codemod.
I think this is being handled in a different branch, to reuse the existing and remove the unused functions.
We are taking the first 5 words from the strings and creating an INTL function and if any function name matches with the existing created we are appending and incrementing the 1,2,3... suffix.

test/intl_suggestors/utils_test.dart
chetansinghchoudhary-wk: Removed unnecessary `expect` function call and updated the test case logic

test/intl_suggestors/utils_test.dart
chetansinghchoudhary-wk: Updated the logic here for `data` variable

test/intl_suggestors/intl_migrator_test.dart
nicbiggs-wk: Ok sounds good to me, thanks for clarifying.

akshaybhardwaj-wk (approved):
LGTM

alanknight-wk (approved):
+10 Tests pass

alanknight-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #200, opened 2022-11-16
INTL-174 : Add nicer logging to codemod dryrun check

akshaybhardwaj-wk:
## Purpose

When repos start adding the codemod dryrun check to their CI, we want them to be able to easily see failures and be able to debug issues. So, let's add some better logging with colors.
## Solution
Create `logging.dart` file in `utill` directory where mention all colors as per `rec.level`

### DIFF
--- a/lib/src/executables/intl_message_migration.dart
+++ b/lib/src/executables/intl_message_migration.dart
@@ -23,11 +23,14 @@ import 'package:file/local.dart';
 import 'package:glob/glob.dart';
 import 'package:glob/list_local_fs.dart';
 import 'package:logging/logging.dart';
+
 import 'package:over_react_codemod/src/intl_suggestors/intl_configs_migrator.dart';
 import 'package:over_react_codemod/src/intl_suggestors/intl_importer.dart';
 import 'package:over_react_codemod/src/intl_suggestors/intl_messages.dart';
 import 'package:over_react_codemod/src/intl_suggestors/intl_migrator.dart';
 import 'package:over_react_codemod/src/util/package_util.dart';
+import 'package:over_react_codemod/src/util/logging.dart';
+
 import 'package:path/path.dart' as p;
 
 import '../util.dart';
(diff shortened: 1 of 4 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

dustinlessard-wf (approved):

dustinlessard-wf:
**### QA Results**
Tested locally:
Screen Shot 2022-11-23 at 1 10 15 PM
Looks great!

Tested as part of the INTL-message Skynet core check:
Screen Shot 2022-11-23 at 1 11 15 PM

Skynet doesn't seem to be honoring the colors.

dustinlessard-wf:
### QA RESULTS:

#### Original:
Screen Shot 2022-11-23 at 1 28 05 PM

#### Tested locally:
Screen Shot 2022-11-23 at 1 10 15 PM
Looks great!

#### Tested as part of the INTL-message Skynet core check:
Screen Shot 2022-11-23 at 1 11 15 PM

Skynet doesn't seem to be honoring the colors.  That's ok for this ticket as that seems out of scope.

When testing locally, we can see how the log message is being output twice, once as the original and again with the new coloring applied.   I don't think we want this.

dustinlessard-wf (dismissed):
The only change I would request is to ensure that log messages are only output once each.

akshaybhardwaj-wk:
> The only change I would request is to ensure that log messages are only output once each.

Will look into this

akshaybhardwaj-wk:
> The only change I would request is to ensure that log messages are only output once each.

I found that the logging text without color is comming from [this code line](https://sourcegraph.plat.workiva.net/github.com/Workiva/dart_codemod/-/blob/lib/src/run_interactive_codemod.dart?L206) from dart_codemod which we are using as dependencies package

brendanburke-wk (approved):

robbecker-wf:
QA+1 CI Passes and output is colored locally. Let's merge it and see if we get colors in our core check or not for real since Micah says the coloring should work 🤷 If we need to revisit to fix the coloring in skynet we can do that in a follow up.
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #241, opened 2023-09-12
INTL-1332: Add import smarts to codemod

chetansinghchoudhary-wk:
### Problem: 
  - Add the ability for the codemod to decide how the intl class import statement will be formatted, ie. package format vs. relative format. It should behave something like this:
      - If the file has only **package** imports, go with a **package** format
      - If the file has only **relative** imports, go with a **relative** format
      - If the file has a mix of both, go with the **package** format
      - If the file has no imports, go with the **package** format
  
 ### Solution:
  - Determine where to insert the import statement. The `_insertionLocationForPackageImport` function is responsible for this.
  - This function inspects the existing import directives in the file and calculates the appropriate insertion point for the `intl` import. 
  - Tracked whether the file has **package** or **relative** imports by iterating through the existing import directives.
  - After determining the insertion location and identifying existing imports, decide on the import format. This is done in the `decideImportFormat` function.
If there are **package** imports and no **relative** imports, it selects the **package** format.
Otherwise, it selects the **relative** format.

### DIFF
--- a/lib/src/intl_suggestors/intl_importer.dart
+++ b/lib/src/intl_suggestors/intl_importer.dart
@@ -48,8 +48,8 @@ Stream<Patch> intlImporter(
 
   final importFormat = decideImportFormat(hasPackageImport, hasRelativeImports);
 
-  final importStatement = importFormat == 'package' ? "import '$intlUri'; "
-                                                    : "import './intl/${projectName}_intl.dart'; ";
+  final importStatement = importFormat == 'package' ? insertInfo.leadingNewlines + "import '$intlUri'; " + insertInfo.trailingNewlines
+                                                    : insertInfo.leadingNewlines + "import './intl/${projectName}_intl.dart'; " + insertInfo.trailingNewlines;
   yield Patch(
       importStatement,
       insertInfo.offset,
--- a/lib/src/intl_suggestors/intl_importer.dart
+++ b/lib/src/intl_suggestors/intl_importer.dart
@@ -31,7 +31,7 @@ Stream<Patch> intlImporter(
   // Parts that have not been generated can show up as `exists = false` but also `isPart = false`,
   // so using the unitResults is a little trickier than using the libraryElement to get it.
   final mainLibraryUnitResult = libraryResult.units.singleWhere((unitResult) =>
-  unitResult.unit.declaredElement ==
+      unitResult.unit.declaredElement ==
       libraryResult.element.definingCompilationUnit);
 
   final needsIntlImport = libraryResult.units
@@ -41,23 +41,30 @@ Stream<Patch> intlImporter(
 
   if (!needsIntlImport) return;
 
-  final intlUri = 'package:${projectName}/src/intl/${projectName}_intl.dart';
-  final currentDirectory = Directory.current;
-  final intlDirectory = path.join(currentDirectory.path, 'lib/src/intl/${projectName}_intl.dart');
-  final relativePathToIntlDir = path.relative(intlDirectory, from: path.current);
-  final levelsUp = List<String>.generate(relativePathToIntlDir.split(path.separator).length, (_) => '..').join('/');
-  final relativeImportPath = '${levelsUp}/src/intl/${projectName}_intl.dart';
+  final intlFilePath = '/src/intl/${projectName}_intl.dart';
+  final intlUri = 'package:${projectName}' + intlFilePath;
+  final intlDirectory = path.join(Directory.current.path, intlFilePath);
+  final relativePathToIntlDir =
+      path.relative(intlDirectory, from: Directory.current.path);
   final insertInfo = _insertionLocationForPackageImport(
       intlUri, mainLibraryUnitResult.unit, mainLibraryUnitResult.lineInfo);
-  final importStatement = insertInfo.usePackageImports  ? packageImport(intlUri, insertInfo) : relativeImport(relativeImportPath, insertInfo);
-  yield Patch(
-      importStatement,
-      insertInfo.offset,
-      insertInfo.offset);
+
+  final importStatement = insertInfo.usePackageImports
+      ? packageImport(intlUri, insertInfo)
+      : relativeImport(relativePathToIntlDir, insertInfo);
+
+  yield Patch(importStatement, insertInfo.offset, insertInfo.offset);
 }
 
-String packageImport(String intlUri, _InsertionLocation insertInfo) => insertInfo.leadingNewlines + "import '$intlUri';" + insertInfo.trailingNewlines;
-String relativeImport(String relativeImportPath, _InsertionLocation insertInfo) =>  insertInfo.leadingNewlines + "import '$relativeImportPath';" + insertInfo.trailingNewlines;
+String packageImport(String intlUri, _InsertionLocation insertInfo) =>
+    insertInfo.leadingNewlines +
+    "import '$intlUri';" +
+    insertInfo.trailingNewlines;
+String relativeImport(
+        String relativeImportPath, _InsertionLocation insertInfo) =>
+    insertInfo.leadingNewlines +
+    "import '$relativeImportPath';" +
+    insertInfo.trailingNewlines;
 
 class _InsertionLocation {
   final int offset;
@@ -66,11 +73,11 @@ class _InsertionLocation {
   final bool usePackageImports;
 
   _InsertionLocation(
-      this.offset, {
-        this.leadingNewlineCount = 0,
-        this.trailingNewlineCount = 0,
-        this.usePackageImports = false,
-      });
+    this.offset, {
+    this.leadingNewlineCount = 0,
+    this.trailingNewlineCount = 0,
+    this.usePackageImports = false,
+  });
 
   String get leadingNewlines => '\n' * leadingNewlineCount;
 
@@ -87,11 +94,11 @@ _InsertionLocation _insertionLocationForPackageImport(
   final firstImport = imports.firstOrNull;
 
   final dartImports =
-  imports.where((i) => i.uriContent?.startsWith('dart:') ?? false);
+      imports.where((i) => i.uriContent?.startsWith('dart:') ?? false);
   final lastDartImport = dartImports.lastOrNull;
 
   final packageImports =
-  imports.where((i) => i.uriContent?.startsWith('package:') ?? false);
+      imports.where((i) => i.uriContent?.startsWith('package:') ?? false);
   final firstPackageImportSortedAfterNewImport = packageImports
       .where((i) => i.uriContent!.compareTo(importUri) > 0)
       .firstOrNull;
@@ -151,6 +158,5 @@ _InsertionLocation _insertionLocationForPackageImport(
       insertAfter ? relativeNode.end : relativeNode.offset,
       leadingNewlineCount: insertAfter ? (inOwnSection ? 2 : 1) : 0,
       trailingNewlineCount: !insertAfter ? (inOwnSection ? 2 : 1) : 0,
-      usePackageImports: hasPackageImports
-  );
+      usePackageImports: hasPackageImports);
 }
--- a/lib/src/intl_suggestors/intl_importer.dart
+++ b/lib/src/intl_suggestors/intl_importer.dart
@@ -153,7 +153,6 @@ _InsertionLocation _insertionLocationForPackageImport(
     return true;
   });
 
-
   return _InsertionLocation(
       insertAfter ? relativeNode.end : relativeNode.offset,
       leadingNewlineCount: insertAfter ? (inOwnSection ? 2 : 1) : 0,
--- a/lib/src/intl_suggestors/intl_importer.dart
+++ b/lib/src/intl_suggestors/intl_importer.dart
@@ -49,7 +49,7 @@ Stream<Patch> intlImporter(
   final importFormat = decideImportFormat(hasPackageImport, hasRelativeImports);
 
   final importStatement = importFormat == 'package' ? insertInfo.leadingNewlines + "import '$intlUri'; " + insertInfo.trailingNewlines
-                                                    : insertInfo.leadingNewlines + "import '../intl/${projectName}_intl.dart'; " + insertInfo.trailingNewlines;
+                                                    : insertInfo.leadingNewlines + "import '../../intl/${projectName}_intl.dart'; " + insertInfo.trailingNewlines;
   yield Patch(
       importStatement,
       insertInfo.offset,
--- a/lib/src/intl_suggestors/intl_importer.dart
+++ b/lib/src/intl_suggestors/intl_importer.dart
@@ -113,7 +113,7 @@ _InsertionLocation _insertionLocationForPackageImport(
   final AstNode relativeNode;
   final bool insertAfter;
   final bool inOwnSection;
-  bool hasPackageImports = true;
+  bool hasOnlyPackageImports = true;
   if (firstPackageImportSortedAfterNewImport != null) {
     relativeNode = firstPackageImportSortedAfterNewImport;
     insertAfter = false;
@@ -141,20 +141,21 @@ _InsertionLocation _insertionLocationForPackageImport(
     // No directive to insert relative to; insert before the first member or
     // at the beginning of the file.
     return _InsertionLocation(unit.declarations.firstOrNull?.offset ?? 0,
-        trailingNewlineCount: 2, usePackageImports: hasPackageImports);
+        trailingNewlineCount: 2, usePackageImports: hasOnlyPackageImports);
   }
 
-  for (final importDirective in imports) {
+  hasOnlyPackageImports = !imports.any((importDirective) {
     final uriContent = importDirective.uri.stringValue;
     if (uriContent != null) {
       final uri = Uri.parse(uriContent);
-      hasPackageImports = !imports.any((importDirective) =>
-          uri != null && uri.scheme != 'package' && uri.scheme != 'dart:');
+      return uri != null && uri.scheme != 'package' && uri.scheme != 'dart:';
     }
-  }
+    return false;
+  });
+
   return _InsertionLocation(
       insertAfter ? relativeNode.end : relativeNode.offset,
       leadingNewlineCount: insertAfter ? (inOwnSection ? 2 : 1) : 0,
       trailingNewlineCount: !insertAfter ? (inOwnSection ? 2 : 1) : 0,
-      usePackageImports: hasPackageImports);
+      usePackageImports: hasOnlyPackageImports);
 }
--- a/lib/src/intl_suggestors/intl_importer.dart
+++ b/lib/src/intl_suggestors/intl_importer.dart
@@ -60,6 +60,7 @@ String packageImport(String intlUri, _InsertionLocation insertInfo) =>
     insertInfo.leadingNewlines +
     "import '$intlUri';" +
     insertInfo.trailingNewlines;
+
 String relativeImport(
         String relativeImportPath, _InsertionLocation insertInfo) =>
     insertInfo.leadingNewlines +
@@ -136,10 +137,11 @@ _InsertionLocation _insertionLocationForPackageImport(
     insertAfter = firstNonImportDirective is LibraryDirective;
     inOwnSection = true;
   } else {
+    bool hasPackageImports = true;
     // No directive to insert relative to; insert before the first member or
     // at the beginning of the file.
     return _InsertionLocation(unit.declarations.firstOrNull?.offset ?? 0,
-        trailingNewlineCount: 2);
+        trailingNewlineCount: 2, usePackageImports: hasPackageImports);
   }
   bool hasPackageImports = true;
 
--- a/lib/src/intl_suggestors/intl_importer.dart
+++ b/lib/src/intl_suggestors/intl_importer.dart
@@ -31,7 +31,7 @@ Stream<Patch> intlImporter(
   // Parts that have not been generated can show up as `exists = false` but also `isPart = false`,
   // so using the unitResults is a little trickier than using the libraryElement to get it.
   final mainLibraryUnitResult = libraryResult.units.singleWhere((unitResult) =>
-      unitResult.unit.declaredElement ==
+  unitResult.unit.declaredElement ==
       libraryResult.element.definingCompilationUnit);
 
   final needsIntlImport = libraryResult.units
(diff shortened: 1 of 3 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -42,10 +42,16 @@ Stream<Patch> intlImporter(
...
   final insertInfo = _insertionLocationForPackageImport(
       intlUri, mainLibraryUnitResult.unit, mainLibraryUnitResult.lineInfo);
+
+  final hasPackageImport = insertInfo.hasPackageImports;
+  final hasRelativeImports = insertInfo.hasRelativeImports;
+
+  final importFormat = decideImportFormat(hasPackageImport, hasRelativeImports);
+
+  final importStatement = importFormat == 'package' ? insertInfo.leadingNewlines + "import '$intlUri'; " + insertInfo.trailingNewlines
+                                                    : insertInfo.leadingNewlines + "import '../../intl/${projectName}_intl.dart'; " + insertInfo.trailingNewlines;
```
alanknight-wk, line 52: I don't think we know for certain that it will be two levels up. A file could be in lib/src, lib/src/project, lib/src/project/components, lib/src/project/components/report_component/redux, etc.

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -124,9 +137,34 @@ _InsertionLocation _insertionLocationForPackageImport(
         trailingNewlineCount: 2);
   }
 
+  imports.forEach((importDirective) {
+    final uriContent = importDirective.uriContent;
+    if(uriContent != null){
+      if(uriContent.startsWith('package:')){
+        hasPackageImports = true;
+      } else if(uriContent.startsWith('../')){
```
alanknight-wk, line 145: A relative import doesn't necessarily start with "..", because it could be importing at the same level, for example
import 'other_file.dart';

You can ask a URI if it's relative or absolute, which I think would be better.

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -124,9 +137,34 @@ _InsertionLocation _insertionLocationForPackageImport(
         trailingNewlineCount: 2);
   }
 
+  imports.forEach((importDirective) {
+    final uriContent = importDirective.uriContent;
```
alanknight-wk, line 141: This needs to only check package: URLS that are for *this* package. e.g.

import 'package:path/path.dart';
import '../foo.dart';

In this case it is using relative imports only.

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -124,9 +137,34 @@ _InsertionLocation _insertionLocationForPackageImport(
         trailingNewlineCount: 2);
   }
 
+  imports.forEach((importDirective) {
```
alanknight-wk, line 140: It's probably preferable to use a regular for-loop for this. They're easier to debug in Dart. Or you could write it as two "any" loops. 

hasPackageImports = imports.any((importDirective) => ...
hasRelativeImports = imports.any(...

lib/src/intl_suggestors/intl_importer.dart
chetansinghchoudhary-wk: Yes, files may be in different packages.

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -42,10 +42,16 @@ Stream<Patch> intlImporter(
 
   final insertInfo = _insertionLocationForPackageImport(
       intlUri, mainLibraryUnitResult.unit, mainLibraryUnitResult.lineInfo);
+
+  final hasPackageImport = insertInfo.hasPackageImports;
+  final hasRelativeImports = insertInfo.hasRelativeImports;
+
+  final importFormat = decideImportFormat(hasPackageImport, hasRelativeImports);
```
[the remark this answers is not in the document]
chetansinghchoudhary-wk: update with `bool` type

lib/src/intl_suggestors/intl_importer.dart
chetansinghchoudhary-wk: Updated with `for` loop

akshaybhardwaj-wk (approved):

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -38,27 +40,48 @@ Stream<Patch> intlImporter(
 
   if (!needsIntlImport) return;
 
-  final intlUri = 'package:${projectName}/src/intl/${projectName}_intl.dart';
+  final currentFilePath = Platform.script.toFilePath();
+  final currentDirectory = path.dirname(currentFilePath);
+
+  final intlUri = 'package:$projectName/src/intl/${projectName}_intl.dart';
+  final relativePathToIntlDir = path.relative(currentDirectory, from: intlUri);
```
alanknight-wk, line 47: Does this (computing a relative path between a URI and a file system path) work in general? I think you could do this more simply as
`intlDirectory = path.join(context.root, 'lib/src/intl/${projectName}_intl.dart');`
`path.relative(intlDirectory, path.current)`
and then you don't need to manually compute relativeImportPath.

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -38,27 +40,48 @@ Stream<Patch> intlImporter(
...
+
+  final levelsUp = List<String>.generate(relativePathToIntlDir.split(path.separator).length, (_) => '..').join('/');
+
+  final relativeImportPath = '$levelsUp/src/intl/${projectName}_intl.dart';
 
   final insertInfo = _insertionLocationForPackageImport(
-      intlUri, mainLibraryUnitResult.unit, mainLibraryUnitResult.lineInfo);
+      relativeImportPath, mainLibraryUnitResult.unit, mainLibraryUnitResult.lineInfo);
```
alanknight-wk, line 54: How does this fix the ordering problem? It seems like this is always finding the insert position as if the import path was relative. Shouldn't you compute the insertInfo after you have the importStatement?

test/intl_suggestors/intl_importer_test.dart
```diff
@@ -139,19 +139,17 @@ void main() {
...
+      
               content() => TestProjectIntl.testString;
             ''',
             isExpectedError: (e) =>
                 isUndefinedIntlError(e) || isFakeUriError(e),
             expectedOutput: /*language=dart*/ '''
-              import 'package:over_react/over_react.dart';
               import 'package:test_project/src/intl/test_project_intl.dart';
-
+              import 'package:over_react/over_react.dart';
```
alanknight-wk, line 150: The point of the rather complicated _insertLocationForPackageImport is that it tries to put it in alphabetical order. This seems like it's just making it so that for the case of package: imports the new import will always comes first (because it computes the position as if it started with '..') and then changes the test cases so that coming first is expected. I think these should go back to the way they were, and the computation should work on the actual form of the URI that we're going to use.

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -157,14 +171,8 @@ _InsertionLocation _insertionLocationForPackageImport(
...
 
-String decideImportFormat(bool hasPackageImports, bool hasRelativeImports) {
-  if (hasPackageImports && !hasRelativeImports) {
-    return 'package';
-  } else if (!hasPackageImports && hasRelativeImports) {
-    return 'relative';
-  } else {
-    return 'package';
-  }
+bool decideImportFormat(bool shouldUsePackageImports, bool shouldUseRelativeImports) {
```
alanknight-wk, line 174: This probably should have a name that indicates what the boolean result means. Like shouldUsePackageImports (but then maybe the parameter by that name needs to change). And a comment to describe the decision.

lib/src/intl_suggestors/intl_importer.dart
alanknight-wk: This is still checking ALL package imports, not just the ones in this package. And the decision process is that we use package imports if there are ANY package imports. So we will not use relative imports if there are imports from any other packages, which is wrong.

test/intl_suggestors/intl_importer_test.dart
```diff
@@ -371,6 +352,39 @@ void main() {
...
+      final importFormat = decideImportFormat(hasPackageImports, hasRelativeImports);
+      expect(importFormat, isFalse);
+    });
+
+    test('when there are both package and relative imports', () {
+      final imports = [
+        "import 'package:over_react/over_react.dart';",
+        "import '../../../src/intl/datatables_intl.dart';",
+      ];
+      final hasPackageImports = imports.any((importStatement) => importStatement.startsWith("import 'package:"));
```
alanknight-wk, line 383: This doesn't seem like the correct test. If we are using relative imports within the package (the only place we can) then we should be using a relative import for the _intl.dart file. Otherwise we might as well not bother doing this, because there are very few files that will only have imports from within their own package.

alanknight-wk (changes requested):

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -38,27 +40,48 @@ Stream<Patch> intlImporter(
 
   if (!needsIntlImport) return;
 
-  final intlUri = 'package:${projectName}/src/intl/${projectName}_intl.dart';
+  final currentFilePath = Platform.script.toFilePath();
```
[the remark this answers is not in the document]
chetansinghchoudhary-wk: Updated as you suggested.

lib/src/intl_suggestors/intl_importer.dart
chetansinghchoudhary-wk: Updated as you suggested.

test/intl_suggestors/intl_importer_test.dart
chetansinghchoudhary-wk: Updated the test cases and

test/intl_suggestors/intl_importer_test.dart
chetansinghchoudhary-wk: The import format is "package" when package and relative imports exist.

lib/src/intl_suggestors/intl_importer.dart
akshaybhardwaj-wk: > Does this (computing a relative path between a URI and a file system path) work in general? I think you could do this more simply as `intlDirectory = path.join(context.root, 'lib/src/intl/${projectName}_intl.dart');` `path.relative(intlDirectory, path.current)` and then you don't need to manually compute relativeImportPath.

I have a one-line `final currentDirectory = Directory.current`; before   `final intlDirectory = path.join(currentDirectory.path, 'lib/src/intl/${projectName}_intl.dart');`

test/intl_suggestors/intl_importer_test.dart
```diff
@@ -60,14 +61,19 @@ void main() {
             input: /*language=dart*/ '''
               import 'package:over_react/over_react.dart';
               content() => TestProjectIntl.testString;
+
```
alanknight-wk, line 64: Probably delete the extra blank lines that were introduced.

test/intl_suggestors/intl_importer_test.dart
```diff
@@ -148,7 +154,7 @@ void main() {
                 isUndefinedIntlError(e) || isFakeUriError(e),
             expectedOutput: /*language=dart*/ '''
               import 'package:over_react/over_react.dart';
-              import 'package:test_project/src/intl/test_project_intl.dart';
+              import '../../../../src/intl/test_project_intl.dart';
```
alanknight-wk, line 157: There should be a test case for a package import in the same package. All these tests produce relative imports.

alanknight-wk (changes requested):

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -39,27 +42,35 @@ Stream<Patch> intlImporter(
   if (!needsIntlImport) return;
 
   final intlUri = 'package:${projectName}/src/intl/${projectName}_intl.dart';
-
+  final currentDirectory = Directory.current;
+  final intlDirectory = path.join(currentDirectory.path, 'lib/src/intl/${projectName}_intl.dart');
+  final relativePathToIntlDir = path.relative(intlDirectory, from: path.current);
```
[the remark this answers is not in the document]
akshaybhardwaj-wk: Got it I have updated the and direct use this `Directory.current.path` instead of taking in variable 
  `final intlDirectory = path.join(Directory.current.path, intlFilePath);`

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -39,27 +42,35 @@ Stream<Patch> intlImporter(
...
+  final levelsUp = List<String>.generate(relativePathToIntlDir.split(path.separator).length, (_) => '..').join('/');
+  final relativeImportPath = '${levelsUp}/src/intl/${projectName}_intl.dart';
   final insertInfo = _insertionLocationForPackageImport(
       intlUri, mainLibraryUnitResult.unit, mainLibraryUnitResult.lineInfo);
+  final importStatement = insertInfo.usePackageImports  ? packageImport(intlUri, insertInfo) : relativeImport(relativeImportPath, insertInfo);
```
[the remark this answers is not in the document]
akshaybhardwaj-wk: code updated

test/intl_suggestors/intl_importer_test.dart
```diff
@@ -44,11 +44,12 @@ void main() {
       test('when there are no other imports', () async {
         await testSuggestor(
           input: /*language=dart*/ '''
+          
               content() => TestProjectIntl.testString;
           ''',
           isExpectedError: isUndefinedIntlError,
           expectedOutput: /*language=dart*/ '''
-              import 'package:test_project/src/intl/test_project_intl.dart';
+              import '../../../../src/intl/test_project_intl.dart';
```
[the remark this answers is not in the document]
akshaybhardwaj-wk: updated the code

dustinlessard-wf (dismissed):

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -38,27 +41,43 @@ Stream<Patch> intlImporter(
 
   if (!needsIntlImport) return;
 
-  final intlUri = 'package:${projectName}/src/intl/${projectName}_intl.dart';
-
+  final intlFilePath = '/src/intl/${projectName}_intl.dart';
+  final intlUri = 'package:${projectName}' + intlFilePath;
```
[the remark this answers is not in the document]
akshaybhardwaj-wk: Code updated

dustinlessard-wf (approved):

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -121,12 +141,23 @@ _InsertionLocation _insertionLocationForPackageImport(
     // No directive to insert relative to; insert before the first member or
     // at the beginning of the file.
     return _InsertionLocation(unit.declarations.firstOrNull?.offset ?? 0,
-        trailingNewlineCount: 2);
+        trailingNewlineCount: 2, usePackageImports: hasPackageImports);
+  }
+
+  for (final importDirective in imports) {
```
alanknight-wk, line 147: This could be more concisely written as something like

bool hasRelativeImports(imports) imports.any((importDirective) => importDirective.uri != null && importDirective.uri.scheme != 'package' && importDirective.uri.scheme != 'dart');

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -121,12 +141,23 @@ _InsertionLocation _insertionLocationForPackageImport(
...
+    final uriContent = importDirective.uri.stringValue;
+    if (uriContent != null) {
+      final uri = Uri.parse(uriContent);
```
alanknight-wk, line 150: I think it's also possible for the scheme to be 'dart:'. I handled that in the comment above.

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -94,6 +113,7 @@ _InsertionLocation _insertionLocationForPackageImport(
   final AstNode relativeNode;
   final bool insertAfter;
   final bool inOwnSection;
+  bool hasPackageImports = true;
```
alanknight-wk, line 116: This name is not clear, as below you're now setting it to false if there are any relative imports at all. Should maybe be hasRelativeImports?

test/intl_suggestors/intl_importer_test.dart
```diff
@@ -65,8 +65,8 @@ void main() {
             expectedOutput: /*language=dart*/ '''
               import 'package:over_react/over_react.dart';
               import 'package:test_project/src/intl/test_project_intl.dart';
-              
               content() => TestProjectIntl.testString;
+
```
alanknight-wk, line 69: added blank line

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -38,27 +41,43 @@ Stream<Patch> intlImporter(
 
   if (!needsIntlImport) return;
 
-  final intlUri = 'package:${projectName}/src/intl/${projectName}_intl.dart';
-
+  final intlFilePath = '/src/intl/${projectName}_intl.dart';
+  final intlUri = 'package:${projectName}' + intlFilePath;
+  final intlDirectory = path.join(Directory.current.path, intlFilePath);
```
alanknight-wk, line 46: These should be using context.root, not Directory.current

lib/src/intl_suggestors/intl_importer.dart
akshaybhardwaj-wk: This `bool` is for enabling true condition where we want imports with the package and if it is not `package` then we are setting it to false.

lib/src/intl_suggestors/intl_importer.dart
akshaybhardwaj-wk: The above code has handled all the cases, for specific `uri.scheme != 'package'` I have done it separately and `hasPackageImports=false `so It working here and above condition as well.  also the `bool usePackageImports` are getting updated in return.

lib/src/intl_suggestors/intl_importer.dart
akshaybhardwaj-wk: Code updated like 
```
 for (final importDirective in imports) {
    final uriContent = importDirective.uri.stringValue;
    if (uriContent != null) {
      final uri = Uri.parse(uriContent);
      hasPackageImports = !imports.any((importDirective) =>
          uri != null && uri.scheme != 'package' && uri.scheme != 'dart:');
    }
  }
```

lib/src/intl_suggestors/intl_importer.dart
alanknight-wk: OK. Maybe it should be hasOnlyPackageImports? Or hasAllPackageImports?

lib/src/intl_suggestors/intl_importer.dart
alanknight-wk: Now you're looping over imports twice. If you have the added line for hasPackageImports =, then you don't need the outer for loop.

alanknight-wk (changes requested):

lib/src/intl_suggestors/intl_importer.dart
akshaybhardwaj-wk: code updated

lib/src/intl_suggestors/intl_importer.dart
```diff
@@ -141,20 +141,21 @@ _InsertionLocation _insertionLocationForPackageImport(
     // No directive to insert relative to; insert before the first member or
     // at the beginning of the file.
     return _InsertionLocation(unit.declarations.firstOrNull?.offset ?? 0,
-        trailingNewlineCount: 2, usePackageImports: hasPackageImports);
+        trailingNewlineCount: 2, usePackageImports: hasOnlyPackageImports);
   }
 
-  for (final importDirective in imports) {
+  hasOnlyPackageImports = !imports.any((importDirective) {
```
alanknight-wk, line 147: Very close. But these lines are not necessary This is taking something that returns a Uri, turning it into a string, then parsing it so you can ask it for its scheme. Just get the uri and check if it's null and check the scheme.

alanknight-wk (changes requested):

lib/src/intl_suggestors/intl_importer.dart
akshaybhardwaj-wk: Initially, we are writing code like [this](https://github.com/Workiva/over_react_codemod/pull/241/commits/eaa80fded45b8034145563c9975131751c3ba8b8#diff-228c7d433e4325d61c63d9f892a715c3dadd37820e9f5a0e2c38ca3677c2ac05R150) 
` final uriContent = importDirective.uriContent;`
but after merge conflicts resolve there is updating in Analyzer 5 `uriContent` is archived and we are using `importDirective.uri.stringValue` 
and if we check scheme with uri like this 

```
hasOnlyPackageImports = !imports.any((importDirective) {
    final uri = importDirective.uri;
    return uri != null && (uri.scheme != 'package' && uri.scheme != 'dart:');
  });

```
then it's giving error `error: The getter 'scheme' isn't defined for the type 'StringLiteral'. (undefined_getter at [over_react_codemod] lib/src/intl_suggestors/intl_importer.dart:158)`

alanknight-wk: Ah, ok. My mistake. I assumed importDirective.uri was returning a Uri, since we needed to convert it to a string. It's a StringLiteral, so this code is good.

alanknight-wk (approved):

alanknight-wk:
QA
The code seems reasonable now, but I ran it for QA on a few repos and it didn't do what I expected.
graph_app - no changes, but they have the codemod in their CI now, so I didn't expect any, and nothing broke.
doc_plat_client - only one change, but it was to a file with no other imports from that package, and it made it a relative import, which doesn't seem right as the default
wdesk_sdk - The oddest one. In lib/src/analytics/shell_analytics.dart there was only one previous import, to another package, it added the intl import as package: But in lib/src/experiences/configs/wdata_experience_config.dart, which did have package: imports in the current package it added it as relative.

akshaybhardwaj-wk:
> lib/src/experiences/configs/wdata_experience_config.dart

check with wdesk_sdk 
1. lib/src/analytics/shell_analytics.dart In this file I am getting package import 
2. lib/src/experiences/configs/wdata_experience_config.dart In this I am getting relative because this file has both `dart:` and `package`  import we have to handle this case and also create a needed test case .

akshaybhardwaj-wk:
> > lib/src/experiences/configs/wdata_experience_config.dart
> 
> check with wdesk_sdk
> 
> 1. lib/src/analytics/shell_analytics.dart In this file I am getting package import
> 2. lib/src/experiences/configs/wdata_experience_config.dart In this I am getting relative because this file has both `dart:` and `package`  import we have to handle this case and also create a needed test case .

handle this scenario and update test case but it's getting failed

alanknight-wk (approved):

alanknight-wk:
QA +1 CI passes

rmconsole-wf (approved):
+1 from RM

PR #253, opened 2023-10-19
Start at fixing the constant migrator's date format detection

alanknight-wk:
## Motivation
  

In attempting to work on something else I put in a constant of the form
const migrateMe = 'migrate me';
into a file, and then wondered why it wasn't getting migrated. Which led me to realize that the boolean expression for avoiding strings like 'MM/dd/yy' was incorrect. So I fixed it, and then realized that it broke a different test, where we screen out strings like 'probably-not' which are, we guess, maybe CSS selectors. But that's not great because it's possible if, um, 'un-likely' to have user-visible strings like that. So if we just fix this logic then it might start picking up other strings we didn't want. And it looks like we don't really actually have any logic for screening those out, though I could have sworn we did. There's some stuff in utils for filtering out pure camelCase alphabetics, but I didn't see this.

Also, the logic here is still questionable, because date formats don't necessarily start with upper case. We'd almost be equally well off just looking for an 'MM/' prefix.

Anyway, I'm leaving this for the moment, but left the code in place for posterity.

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist

### DIFF
--- a/lib/src/intl_suggestors/intl_migrator.dart
+++ b/lib/src/intl_suggestors/intl_migrator.dart
@@ -119,18 +119,19 @@ class ConstantStringMigrator extends GeneralizingAstVisitor
       if (string.toUpperCase() == string) return;
       // Is the first character uppercase, excluding strings that start with two of the same
       // uppercase character, which has a good chance of being a date format (e.g. 'MM/dd/YYYY').
-      if (firstLetter != secondLetter &&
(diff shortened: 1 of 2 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

PR #254, opened 2023-10-19
Quick fix experiment

alanknight-wk:
## Motivation
  

The codemod is heavy, and designed for large migrations a package at a time. We'd like to be able to re-use the logic, but do it in a faster and much smaller way. This is an experiment in invoking something quickly that can make an individual change.

My thought is to write a special-purpose migrator that takes a file name and a character position. So if we give it a character position anywhere within a string it migrates just that string and exits.

Right now, all this does is run just the constant migrator on a single file. If it's already been compiled, then it runs in sub-second time. But that's not doing a `ddev format` on the _intl.dart file. So we could accept that and leave it unformatted, or we could make the IDE extension run the format on save operation, which is generally faster, or something else. I'm thinking something like

`dart pub global run over_react_codemod:intl_quick --character-position=12345 --file=lib/src/stuff/thing.dart`

So this seems reasonably promising. The migrators can run without a resolved analysis, and do it pretty quickly.

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist

### DIFF
--- a/lib/src/executables/intl_quick_migration.dart
+++ b/lib/src/executables/intl_quick_migration.dart
@@ -1,0 +1,104 @@
+// Copyright 2021 Workiva Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:io';
+
+import 'package:args/args.dart';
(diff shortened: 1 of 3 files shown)


lib/src/executables/intl_quick_migration.dart
```diff
@@ -0,0 +1,104 @@
...
+  messages.write(force: false);
+  // This will leave the intl.dart file unformatted, but that takes too long, so we'll just leave it out.
+}
+
+Future<int> runMigrators(List<String> packageDartPaths,
+    List<String> codemodArgs, IntlMessages messages, String packageName) async {
+  // final intlPropMigrator = IntlMigrator(messages.className, messages);
+  final constantStringMigrator =
+      ConstantStringMigrator(messages.className, messages);
+  // final importMigrator = (FileContext context) =>
```
alanknight-wk, line 91: Should probably run the importer too, so unwind the change to run only a single migrator. But we shouldn't (I think) need the change that runs them all sequentially. Or we could let the IDE Extension still do the import.

aviary-wf:
## Security Insights

## Action Items
---

PR #259, opened 2023-10-25
INTL-1685: improve failure message

dustinlessard-wf:
## Motivation

When engineers see a CI check failure, they are not provided any indication of next steps.

> 1 change(s) needed.

## Changes

- modified failure message to include instructions on how to run the INTLcodemod locally

Screenshot 2023-10-26 at 10 07 04 AM

#### Release Notes
modify failure message to include instructions on how to run the INTL codemod locally

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist

### DIFF
--- a/lib/src/executables/intl_message_migration.dart
+++ b/lib/src/executables/intl_message_migration.dart
@@ -22,6 +22,7 @@ import 'package:file/file.dart';
 import 'package:file/local.dart';
 import 'package:glob/glob.dart';
 import 'package:glob/list_local_fs.dart';
+import 'package:io/ansi.dart' as ansi;
 import 'package:logging/logging.dart';
 import 'package:over_react_codemod/src/intl_suggestors/intl_configs_migrator.dart';
 import 'package:over_react_codemod/src/intl_suggestors/intl_importer.dart';
@@ -210,7 +211,7 @@ void main(List<String> args) async {
 }
 
 void printInBlue(String text) {
-  print('\x1B[34m$text\x1B[0m');
+  print(ansi.blue.wrap(text));
 }
 
 void printUsage() {
--- a/lib/src/executables/intl_message_migration.dart
+++ b/lib/src/executables/intl_message_migration.dart
@@ -199,6 +199,18 @@ void main(List<String> args) async {
     await migratePackage(
         package, packageNameLookup, processedPackages, codemodArgs, dartPaths);
   }
+
+  if (exitCode != 0) {
+    printInBlue(
+        'To resolve these changes, please execute the codemod locally by running :');
+    printInBlue('dart pub global activate --overwrite over_react_codemod');
+    printInBlue(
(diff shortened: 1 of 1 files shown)


lib/src/executables/intl_message_migration.dart
```diff
@@ -199,6 +199,18 @@ void main(List<String> args) async {
...
+    printInBlue(
+        'To resolve these changes, please execute the codemod locally by running :');
+    printInBlue('dart pub global activate --overwrite over_react_codemod');
+    printInBlue(
+        'dart pub global run over_react_codemod:intl_message_migration --migrate-constants --prune-unused');
+  }
+}
+
+void printInBlue(String text) {
+  print('\x1B[34m$text\x1B[0m');
```
alanknight-wk, line 213: Surely there's something better than hard-coding the escape sequences. Looks like there might be other things we're using, e.g. https://sourcegraph.plat.workiva.net/github.com/Workiva/over_react/-/blob/tools/analyzer_plugin/tool/init_local_dev.dart?L60:15

lib/src/executables/intl_message_migration.dart
```diff
@@ -199,6 +199,18 @@ void main(List<String> args) async {
     await migratePackage(
         package, packageNameLookup, processedPackages, codemodArgs, dartPaths);
   }
+
+  if (exitCode != 0) {
```
alanknight-wk, line 203: The impression I had from the discussion was that it would be nice if it could actually list the problem code locations/strings. Some people have expressed the desire to do the changes manually using the IDE extensions rather than running the codemod.

lib/src/executables/intl_message_migration.dart
dustinlessard-wf: Agreed, will change to using the io library.

lib/src/executables/intl_message_migration.dart
dustinlessard-wf: I share that impression from one of our conversations, yet in another conversation, there was this question: "
How do I find the change that needs to be made in this without reading 7000 lines?"  If we append to the existing logs in the CI check, we're adding to that large quantity of lines, where as if we append the 3 lines at the end and encourage folks to use the codemod locally, they will be prompted with each change, and can step through the changes needed.

I think for now, we should take an incremental step forward and see if this will meet most people's needs, then circle back with folks with whether we need to do more.  Additionally, I'd like to talk with FEA about reducing the amount of low-value codemod-related logs firts, before we add much more.

Thoughts?

alanknight-wk (approved):

alanknight-wk (approved):

alanknight-wk:
@Workiva/release-management-p

dustinlessard-wf:
QA+1

rmconsole-wf (approved):
+1 from RM

PR #266, opened 2023-11-21
INTL-1581 : Add wiki link to exceptions Codemod Crash Information in our Wiki

akshaybhardwaj-wk:
## Problem
  
When we run `codemod` and fund any exception then we have no link in the guideline link in the error stack. INTL team created a document on wiki Codemod Exception [Handling](https://wiki.atl.workiva.net/display/FEF/Codemod+Exception+Handling) 
## Solution
  
Add the Link of wiki [page]( https://wiki.atl.workiva.net/display/FEF/Codemod+Exception+Handling) in the error log. Like 
`logger.warning('Failed to read file: ${file.path} - More details: https://wiki.atl.workiva.net/display/FEF/Codemod+Exception+Handling', e, stack);`

### DIFF
--- a/lib/src/executables/dart2_9_upgrade.dart
+++ b/lib/src/executables/dart2_9_upgrade.dart
@@ -162,7 +162,10 @@ extension on RecursiveAstVisitor {
       try {
         sourceText = file.readAsStringSync();
       } catch (e, stack) {
-        logger.warning('Failed to read file: ${file.path} - More details: https://wiki.atl.workiva.net/display/FEF/Codemod+Exception+Handling', e, stack);
+        logger.warning(
+            'Failed to read file: ${file.path} - More details: https://wiki.atl.workiva.net/display/FEF/Codemod+Exception+Handling',
+            e,
+            stack);
         continue;
       }
 
--- a/lib/src/executables/dart2_9_upgrade.dart
+++ b/lib/src/executables/dart2_9_upgrade.dart
@@ -162,7 +162,7 @@ extension on RecursiveAstVisitor {
       try {
         sourceText = file.readAsStringSync();
       } catch (e, stack) {
-        logger.warning('Failed to read file: ${file.path}', e, stack);
+        logger.warning('Failed to read file: ${file.path} - More details: https://wiki.atl.workiva.net/display/FEF/Codemod+Exception+Handling', e, stack);
         continue;
       }
 
--- a/lib/src/executables/dependency_validator_ignore.dart
+++ b/lib/src/executables/dependency_validator_ignore.dart
@@ -65,10 +65,10 @@ void main(List<String> args) async {
       print('[${rec.level}] ${rec.message}');
     }
     if (rec.error != null) {
-      print(rec.error);
+      print('Error: ${rec.error} \n more details: https://wiki.atl.workiva.net/display/FEF/Codemod+Exception+Handling');
     }
     if (rec.stackTrace != null) {
-      print(rec.stackTrace);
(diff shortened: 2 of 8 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

nicbiggs-wk (changes requested):
Unfortunately, we have moved all of our codemod stuff into the `intl_tools` repo as a sub-package (`pkg/intl_codemod`) so that is where these changes will need to happen. We both moved the `intl_suggestors` stuff as is and copied over some other files that were needed for them, with unused things removed. So, I believe the simplest way to do this would be to:
1. For each `intl_suggestor` file, just copy them over and keep the changes you added. You will likely need to make sure the import statements refer to `intl_codemod` instead of `over_react_codemod` so watch out for that.
2. For every other file, make sure they were already copied over before making the change. We only want your changes made to the files in there, no adding anything else.
If the changes to other files that weren't copied over need to happen, I'm not sure what the best approach for those are.

nicbiggs-wk:
One other comment, I'd recommend storing the URL link to the wiki into a constant variable, and then using the variable in the log warnings. That way if we ever need to change the referenced page, it'll be a lot easier to only change it in the one place. 
For example:
`const exceptionUrl = 'https://wiki.atl.workiva.net/display/FEF/Codemod+Exception+Handling';`

You might even consider having a constant for the whole message and just adding that to the log messages.
For example:
`const exceptionMessage = ' more details: https://wiki.atl.workiva.net/display/FEF/Codemod+Exception+Handling';`

akshaybhardwaj-wk:
> Unfortunately, we have moved all of our codemod stuff into the `intl_tools` repo as a sub-package (`pkg/intl_codemod`) so that is where these changes will need to happen. We both moved the `intl_suggestors` stuff as is and copied over some other files that were needed for them, with unused things removed. So, I believe the simplest way to do this would be to:
> 
> 1. For each `intl_suggestor` file, just copy them over and keep the changes you added. You will likely need to make sure the import statements refer to `intl_codemod` instead of `over_react_codemod` so watch out for that.
> 2. For every other file, make sure they were already copied over before making the change. We only want your changes made to the files in there, no adding anything else.
>    If the changes to other files that weren't copied over need to happen, I'm not sure what the best approach for those are.

@nicbiggs-wk 
Currently, we have only migrated two executables: **message_migration** and **sort**. Within these components, we have addressed **dart2_9_upgrade** and **dependency_validator_ignore**, where I have included a wiki link to the exception handling. In the **intl_suggestors** and other files, I have simply applied a null safe cast_nullable_to_non_nullable ignore the comment. 

Do we need to move both of them also?

To transfer **dart2_9_upgrade** and **dependency_validator_ignore** to the **INTL_tool**, we need to create a separate ticket. Once this task is completed, we will close the current pull request ([PR](https://github.com/Workiva/over_react_codemod/pull/266)) and generate a new one specifically for the INTL_tool.

akshaybhardwaj-wk:
> One other comment, I'd recommend storing the URL link to the wiki into a constant variable, and then using the variable in the log warnings. That way if we ever need to change the referenced page, it'll be a lot easier to only change it in the one place. For example: `const exceptionUrl = 'https://wiki.atl.workiva.net/display/FEF/Codemod+Exception+Handling';`
> 
> You might even consider having a constant for the whole message and just adding that to the log messages. For example: `const exceptionMessage = ' more details: https://wiki.atl.workiva.net/display/FEF/Codemod+Exception+Handling';`

I will create a separate  file `/src/contants.dart` this  will place the constant there and import this file in the file where we need the constants

akshaybhardwaj-wk:
@nicbiggs-wk closing this PR and moving the code related to `intl_message_migration` and suggested changes related to constants into `int_tools`

PR #268, opened 2023-12-13
FED-1919 Codemod to add required flux actions / store prop(s)

aaronlademann-wf:
## Motivation
The null safety over_react major release will make a breaking change in which `FluxPropsMixin`'s `store` and `actions` props are required, so that they can be made non-nullable.

Making `store` required shouldn't affect most usages, but `actions` will definitely affect some consumers.

While the `actions` prop will be present in usages in most cases, there are usages that only wire up a component to redraw on the store, because the actions aren't used. 

## Changes
1. Create a `RequiredFluxProps()` codemod class that patches invoked and un-invoked builders that mix-in `FluxUiPropsMixin` when either the `actions` or `store` prop(s) are missing.
    - Adds the missing setter with a `null` value unless an in-scope var with matching type is found.
    - If an in-scope var is found with a matching type - that var is set as the value.
1. Write tests for all conceivable patch cases

## QA
1. Verify the changes produced by [this batch](https://github.com/Workiva/sourcegraph-batch-changes/pull/325) that executes this codemod.


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
```diff
@@ -0,0 +1,230 @@
...
+  if (inScopePropName != null) {
+    return '${componentScopePropDetector.propsName}.${inScopePropName}';
+  }
+
+  return inScopeVarName;
+}
+
+bool _isFnComponentDeclaration(Expression? varInitializer) =>
+    varInitializer is MethodInvocation &&
+    varInitializer.methodName.name.startsWith('uiF');
```
aaronlademann-wf, line 150: Should account for both `uiFunction` and `uiForwardRef`

lib/src/intl_suggestors/intl_migrator.dart
```diff
@@ -108,7 +108,7 @@ class ConstantStringMigrator extends GeneralizingAstVisitor
     if (node.isConst &&
         node.initializer != null &&
         node.initializer is SimpleStringLiteral) {
-      SimpleStringLiteral literal = node.initializer as SimpleStringLiteral;
+      SimpleStringLiteral literal = node.initializer! as SimpleStringLiteral;
```
aaronlademann-wf, line 111: There were lints causing CI failures in a few misc. files that are unrelated to the scope of this PR.

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
```diff
@@ -0,0 +1,228 @@
...
+  node.thisOrAncestorOfType<CompilationUnit>()?.accept(inScopeVariableDetector);
+  // Find vars declared in top-level fns (like `main()`)
+  node
+      .thisOrAncestorOfType<BlockFunctionBody>()
+      ?.visitChildren(inScopeVariableDetector);
+
+  final inScopeVarName = inScopeVariableDetector.found
+      .firstWhereOrNull((v) {
+        final maybeMatchingType = v.declaredElement?.type;
+        return maybeMatchingType?.element?.name == type?.element?.name;
```
greglittlefield-wf, line 111: Instead of just matching the name, I think we could use `TypeSystem`, which would match any subtypes, and prevent any false positives where in scope variables are different types with the same name.
```dart
  // Usually we'd grab typeSystem from the ResolvedUnitResult, but we don't have access to that
  // in this class, so just get it from the compilation unit.
  final typeSystem = (node.root as CompilationUnitElement).library.typeSystem;
  bool isMatchingType(DartType? maybeMatchingType) =>
      maybeMatchingType != null &&
      typeSystem.isAssignableTo(maybeMatchingType, type);

  final inScopeVarName = inScopeVariableDetector.found
      .firstWhereOrNull((v) => isMatchingType(v.declaredElement?.type))
      ?.declaredElement
      ?.name;
```

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
```diff
@@ -0,0 +1,228 @@
...
+    if (_isFnComponentDeclaration(node.initializer)) return;
+
+    if (node.declaredElement != null) {
+      found.add(node);
+    }
+  }
+}
+
+/// A visitor to detect store/actions values in a props class (supports both class and fn components)
+class _ComponentScopeFluxPropsDetector extends RecursiveAstVisitor<void> {
```
greglittlefield-wf, line 168: #nit it doesn't look like this class is really being used as a visitor, since it only processes the nodes we directly pass to it.

We could make it a non-visitor and it would function just the same:
```diff
diff --git a/lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart b/lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
index 9d84eb9..3617886 100644
--- a/lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
+++ b/lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
@@ -115,13 +115,11 @@ String? _getNameOfVarOrFieldInScopeWithType(AstNode node, DartType? type) {
 
   final componentScopePropDetector = _ComponentScopeFluxPropsDetector();
   // Find actions/store in props of class components
-  node
-      .thisOrAncestorOfType<ClassDeclaration>()
-      ?.accept(componentScopePropDetector);
+  componentScopePropDetector.handlePotentialClassComponent(
+      node.thisOrAncestorOfType<ClassDeclaration>());
   // Find actions/store in props of fn components
-  node
-      .thisOrAncestorOfType<MethodInvocation>()
-      ?.accept(componentScopePropDetector);
+  componentScopePropDetector.handlePotentialFunctionComponent(
+      node.thisOrAncestorOfType<MethodInvocation>());
 
   final inScopePropName =
       componentScopePropDetector.found.firstWhereOrNull((el) {
@@ -165,8 +163,9 @@ class _InScopeVarDetector extends RecursiveAstVisitor<void> {
 }
 
 /// A visitor to detect store/actions values in a props class (supports both class and fn components)
-class _ComponentScopeFluxPropsDetector extends RecursiveAstVisitor<void> {
+class _ComponentScopeFluxPropsDetector {
   final Map<PropertyAccessorElement, DartType> _foundWithMappedTypes;
+
   List<PropertyAccessorElement> get found =>
       _foundWithMappedTypes.keys.toList();
 
@@ -201,8 +200,8 @@ class _ComponentScopeFluxPropsDetector extends RecursiveAstVisitor<void> {
   }
 
   /// Visit function components
-  @override
-  void visitMethodInvocation(MethodInvocation node) {
+  void handlePotentialFunctionComponent(MethodInvocation? node) {
+    if (node == null) return;
     if (!_isFnComponentDeclaration(node)) return;
 
     final nodeType = node.staticType;
@@ -219,8 +218,8 @@ class _ComponentScopeFluxPropsDetector extends RecursiveAstVisitor<void> {
   }
 
   /// Visit composite (class) components
-  @override
-  void visitClassDeclaration(ClassDeclaration node) {
+  void handlePotentialClassComponent(ClassDeclaration? node) {
+    if (node == null) return;
     final elWithProps =
         node.declaredElement?.supertype?.typeArguments.singleOrNull?.element;
     _lookForFluxStoreAndActionsInPropsClass(elWithProps);
```

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
```diff
@@ -0,0 +1,228 @@
...
+          'Could not get resolved result for "${context.relativePath}"');
+    }
+    _result!.unit.accept(this);
+  }
+}
+
+String? _getNameOfVarOrFieldInScopeWithType(AstNode node, DartType? type) {
+  final inScopeVariableDetector = _InScopeVarDetector();
+  // Find top level vars
+  node.thisOrAncestorOfType<CompilationUnit>()?.accept(inScopeVariableDetector);
```
greglittlefield-wf, line 102: It looks like this also finds out of scope variables, since it's starting at the compilation unit and recursing through everything (except function components).

For instance:
```dart
class MyComponent extends FluxUiComponent2... {
  render() {
    MyStore store; // Would pick this up
  }
}

void someOtherFunction() {
  Store store; // Would also pick this up
}

class MyModuleComponents {
  void someMethod() {
    MyStore store; // Would also pick this up
  }
}
```

This also only handles variable declarations, and not arguments.

I don't think Dart has any APIs for getting all the in-scope names, so we'll definitely just have to keep approximating this. But I think we should probably tweak this implementation to be more conservative.

How about something like this:
```diff
diff --git a/lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart b/lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
index 9d84eb9..694a83c 100644
--- a/lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
+++ b/lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
@@ -18,6 +18,7 @@ import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
 import 'package:collection/collection.dart';
+import 'package:over_react_codemod/src/util.dart';
 import 'package:over_react_codemod/src/util/class_suggestor.dart';
 import 'package:over_react_codemod/src/util/offset_util.dart';
 
@@ -96,22 +97,46 @@ class RequiredFluxProps extends RecursiveAstVisitor with ClassSuggestor {
   }
 }
 
+class InScopeVariable {
+  final String name;
+  final DartType? type;
+
+  InScopeVariable(this.name, this.type);
+}
+
 String? _getNameOfVarOrFieldInScopeWithType(AstNode node, DartType? type) {
-  final inScopeVariableDetector = _InScopeVarDetector();
-  // Find top level vars
-  node.thisOrAncestorOfType<CompilationUnit>()?.accept(inScopeVariableDetector);
-  // Find vars declared in top-level fns (like `main()`)
-  node
-      .thisOrAncestorOfType<BlockFunctionBody>()
-      ?.visitChildren(inScopeVariableDetector);
+  final mostInScopeVariables = node.ancestors.expand((ancestor) sync* {
+    if (ancestor is FunctionDeclaration) {
+      // Function arguments
+      final element = ancestor.declaredElement;
+      if (element != null) {
+        yield* element.parameters.map((p) => InScopeVariable(p.name, p.type));
+      }
+    } else if (ancestor is Block) {
+      // Variables declared in the block (function body, if/else block, etc.)
+      yield* ancestor.statements
+          .whereType<VariableDeclarationStatement>()
+          .expand((d) => d.variables.variables)
+          .map((v) => InScopeVariable(v.name.lexeme, v.declaredElement?.type));
+    } else if (ancestor is ClassDeclaration) {
+      // Class fields
+      final element = ancestor.declaredElement;
+      if (element != null) {
+        yield* element.fields.map((f) => InScopeVariable(f.name, f.type));
+      }
+    } else if (ancestor is CompilationUnit) {
+      // Top-level variables
+      yield* ancestor.declarations
+          .whereType<TopLevelVariableDeclaration>()
+          .expand((d) => d.variables.variables)
+          .map((v) => InScopeVariable(v.name.lexeme, v.declaredElement?.type));
+    }
+  });
 
-  final inScopeVarName = inScopeVariableDetector.found
-      .firstWhereOrNull((v) {
-        final maybeMatchingType = v.declaredElement?.type;
-        return maybeMatchingType?.element?.name == type?.element?.name;
-      })
-      ?.declaredElement
-      ?.name;
+  final inScopeVarName = mostInScopeVariables.firstWhereOrNull((v) {
+    final maybeMatchingType = v.type;
+    return maybeMatchingType?.element?.name == type?.element?.name;
+  })?.name;
 
   final componentScopePropDetector = _ComponentScopeFluxPropsDetector();
   // Find actions/store in props of class components
@@ -147,26 +172,10 @@ bool _isFnComponentDeclaration(Expression? varInitializer) =>
     varInitializer is MethodInvocation &&
     varInitializer.methodName.name.startsWith('uiF');
 
-/// A visitor to detect in-scope store/actions variables (top-level and block function scopes)
-class _InScopeVarDetector extends RecursiveAstVisitor<void> {
-  final List<VariableDeclaration> found;
-
-  _InScopeVarDetector() : found = [];
-
-  @override
-  visitVariableDeclaration(VariableDeclaration node) {
-    // Don't visit function component declarations here since we visit them using the _ComponentScopeFluxPropsDetector
-    if (_isFnComponentDeclaration(node.initializer)) return;
-
-    if (node.declaredElement != null) {
-      found.add(node);
-    }
-  }
-}
-
```

greglittlefield-wf (changes requested):
Made a pass at everything except tests; will take a look at those after lunch

test/dart3_suggestors/null_safety_prep/required_flux_props_test.dart
```diff
@@ -0,0 +1,1160 @@
...
+        resolvedContext: resolvedContext,
+      );
+    });
+
+    test(
+        'leaves builders alone if they don\'t use FluxUiPropsMixin, '
+        'even if they have props named store/actions', () async {
+      await testSuggestor(
+        isExpectedError: (err) =>
+            err.message.contains(RegExp(r'theStore|theActions')),
```
greglittlefield-wf, line 45: Is this error to ignore `unused_local_variable`? Could we test for that error message instead, to help avoid accidentally ignoring other errors related to those variables that could be indicative of a bad test setup?

test/dart3_suggestors/null_safety_prep/required_flux_props_test.dart
```diff
@@ -0,0 +1,1160 @@
...
+class FooProps = UiProps with FluxUiPropsMixin<${actionsName ?? 'Null'}, ${storeName ?? 'Null'}>;
+
+class FooComponent extends FluxUiComponent2<FooProps> {
+  @override
+  render() => null;
+}
+
+UiFactory<NotFooProps> NotFoo = castUiFactory(_\$NotFoo); // ignore: undefined_identifier
+
+// FIXME (adl): Do we need to have a mixin that has store/actions with the correct type? (See FIXME in impl regarding out-of-scope vars being used)
```
greglittlefield-wf, line 1148: Should this fixme be addressed?

test/dart3_suggestors/null_safety_prep/required_flux_props_test.dart
```diff
@@ -0,0 +1,1160 @@
...
+                      ..actions = theActions
+                      ..store = theStore
+                    ''')}
+                  }
+                '''),
+              );
+            });
+          });
+
+          group('when an actions var is available in block function scope', () {
```
greglittlefield-wf, line 154: (With the updates I suggested) I think it'd be good to add a couple test cases for function parameters and class fields.

greglittlefield-wf (changes requested):
Done with second pass, just a couple small comments on tests; overall looks great!

test/dart3_suggestors/null_safety_prep/required_flux_props_test.dart
aaronlademann-wf: It already has been (via class / function props consumption tests)

test/dart3_suggestors/null_safety_prep/required_flux_props_test.dart
aaronlademann-wf: Added in https://github.com/Workiva/over_react_codemod/pull/268/commits/aea20f00a1cad1436a176e585a6341cc64c48800

test/dart3_suggestors/null_safety_prep/required_flux_props_test.dart
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react_codemod/pull/268/commits/6a3aafc1a64e15b27034b8f6221bcab2ff4525f8

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react_codemod/pull/268/commits/aea20f00a1cad1436a176e585a6341cc64c48800

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react_codemod/pull/268/commits/619910d7e5bd4bf060862cac0d95a97ad8b2e327

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react_codemod/pull/268/commits/580b82cd2dbf752e89dba8a936ee6a047091274b

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
```diff
@@ -0,0 +1,228 @@
...
+
+    DartType? fluxActionsType;
+    DartType? fluxStoreType;
+    final cascadeWriteEl = node.staticType?.element;
+    if (cascadeWriteEl is ClassElement) {
+      final maybeFluxUiPropsMixin = cascadeWriteEl.mixins
+          .singleWhereOrNull((e) => e.element.name == fluxPropsMixinName);
+      writesToFluxUiProps = maybeFluxUiPropsMixin != null;
+      fluxActionsType = maybeFluxUiPropsMixin?.typeArguments[0];
+      fluxStoreType = maybeFluxUiPropsMixin?.typeArguments[1];
```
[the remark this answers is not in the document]
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react_codemod/pull/268/commits/b2597cacc6789fbfaffc77f571b9659c561bfb92

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
```diff
@@ -42,6 +42,15 @@ class RequiredFluxProps extends RecursiveAstVisitor with ClassSuggestor {
   visitCascadeExpression(CascadeExpression node) {
     final cascadeWriteEl = node.staticType?.element;
     if (cascadeWriteEl is! ClassElement) return;
+    final isReturnedAsDefaultProps = node.ancestors
+            .whereType<MethodDeclaration>()
+            .firstOrNull
+            ?.name
+            .value()
+            .toString()
```
greglittlefield-wf, line 50: #nit This can be just
```suggestion
            .lexeme
```

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
```diff
@@ -42,6 +42,15 @@ class RequiredFluxProps extends RecursiveAstVisitor with ClassSuggestor {
...
+            .contains(RegExp(r'getDefaultProps|defaultProps')) ==
+        true;
```
greglittlefield-wf, line 52: #nit
https://dart.dev/effective-dart/usage#dont-use-true-or-false-in-equality-operations
```suggestion
            .contains(RegExp(r'getDefaultProps|defaultProps')) ??
        false;
```

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
```diff
@@ -77,7 +86,15 @@ class RequiredFluxProps extends RecursiveAstVisitor with ClassSuggestor {
...
-    final offset = context.sourceFile.getOffsetOfLineAfter(node.target.offset);
+    final sf = context.sourceFile;
+    final targetLineOffset = sf.getOffset(sf.getLine(node.target.offset));
+    int offset;
+    if (targetLineOffset == sf.getOffset(sf.getLine(node.target.end))) {
+      // Cascade on a single line / same line as the target, add the new setter(s) before the semicolon
+      offset = node.target.end;
+    } else {
+      offset = sf.getOffsetOfLineAfter(node.target.offset);
+    }
```
greglittlefield-wf, line 97: #nit why are we trying to conditionally insert it on a new line? This will all end up getting formatted anyways
```suggestion
    final offset = node.target.end;
```

greglittlefield-wf (dismissed):
+1 with some #nits

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react_codemod/pull/268/commits/d36a65610c17e2d5768b344bc5ca0038efe5206d

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react_codemod/pull/268/commits/d36a65610c17e2d5768b344bc5ca0038efe5206d

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react_codemod/pull/268/commits/d36a65610c17e2d5768b344bc5ca0038efe5206d

greglittlefield-wf (approved):

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
```diff
@@ -0,0 +1,247 @@
...
+
+  // Usually we'd grab typeSystem from the ResolvedUnitResult, but we don't have access to that
+  // in this class, so just get it from the compilation unit.
+  final typeSystem =
+      (node.root as CompilationUnit).declaredElement!.library.typeSystem;
+  bool isMatchingType(DartType? maybeMatchingType) =>
+      maybeMatchingType != null &&
+      maybeMatchingType is! DynamicType &&
+      type is! DynamicType &&
+      typeSystem.isAssignableTo(maybeMatchingType, type);
```
greglittlefield-wf, line 147: I noticed some places where store was getting assigned to some unexpected values (for example, [here](https://sourcegraph.plat.workiva.net/organizations/Workiva/batch-changes/apply/QmF0Y2hTcGVjOiJkYTRmZDc0OS0zYjNlLTRjMDEtYTU1My1lYzQxYjhmM2UyN2Yi?visible=33#diff-e8eeaa246874b37275c86779da0b0843)). 

It looks like for at least one (`SidePanelContents`), it's because the TStore is `Null`. Perhaps we should have another check for `type.isDartCoreNull`?

Come to think of it, we could probably just short-circuit in this method before we even start to search:
```dart
String? _getNameOfVarOrFieldInScopeWithType(AstNode node, DartType type) {
  if (type is DynamicType || type.isDartCoreNull) return null;
  // ...
```

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
```diff
@@ -0,0 +1,247 @@
...
+            .firstOrNull
+            ?.name
+            .lexeme
+            .contains(RegExp(r'getDefaultProps|defaultProps')) ??
+        false;
+    if (isReturnedAsDefaultProps) return;
+
+    final maybeFluxUiPropsMixin = cascadeWriteEl.mixins
+        .singleWhereOrNull((e) => e.element.name == fluxPropsMixinName);
+    if (maybeFluxUiPropsMixin == null) return;
```
greglittlefield-wf, line 56: I noticed a lot of cases of `PanelTitle` and `PanelTitleV2`, which happen to mix in FluxUiPropsMixin, but doesn't actually use those props. 

Adding/props/store to these is pretty noisy, and as part of FED-2015 I think we'll suppress required prop errors for them, so I feel like it would be worth also making an exception for them. 

I think you should be able to do that with this:
```dart
    const typesToIgnore = {'_PanelTitleProps', 'PanelTitleProps', 'PanelTitleV2Props'};
    if (typesToIgnore.contains(cascadeWriteEl.name)) {
      return;
    }
```

greglittlefield-wf (changes requested):
Couple more comments after reviewing the test batch

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
```diff
@@ -0,0 +1,247 @@
...
+    });
+    var actionsAssigned = cascadingAssignments.any((cascade) {
+      final lhs = cascade.leftHandSide;
+      return lhs is PropertyAccess && lhs.propertyName.name == 'actions';
+    });
+
+    if (!storeAssigned) {
+      storeAssigned = true;
+      final storeValue =
+          _getNameOfVarOrFieldInScopeWithType(node, fluxStoreType) ?? 'null';
```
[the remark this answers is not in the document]
aaronlademann-wf: Fixed / added tests in https://github.com/Workiva/over_react_codemod/pull/268/commits/e8300cb529ced7eabccbc6e54001cd4d6f0d6941

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
aaronlademann-wf: Fixed / added tests in https://github.com/Workiva/over_react_codemod/pull/268/commits/a603afe57c07b05115943f49fa5ad000708a2d84

lib/src/dart3_suggestors/null_safety_prep/required_flux_props.dart
aaronlademann-wf: Fixed / added tests in https://github.com/Workiva/over_react_codemod/pull/268/commits/0f3a754e5c6a5c4913b6c000cc8910a2e6c21c1a

greglittlefield-wf (dismissed):

greglittlefield-wf (approved):
+10

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #271, opened 2024-02-27
Remove web_skin_dart all.dart export usages - manual fixup to use component-specific imports

sourcegraph-wk:
web_skin_dart has exports for individual components and an "all" export that has all components.  If everywhere uses the individual component exports, then the compiler will only have to consider the component code it needs, resulting in faster build times.  But, this only works if everywhere is using the individual component exports.
 
 This batch removes the all.dart import.  It will require manual fixup using an IDE's fix suggestions to import the component exports.  If you would like to help us fan out and do this simple manual work to get rid of all analyzer errors, that would be fantastic.  Once CI is green, you can review and merge it.  Be careful to only import `component2` exports.  If you import the old component1 symbols, your components may break or regress.
 
[_Created by Sourcegraph batch change `Workiva/web_skin_dart_all_import`._](https://sourcegraph.plat.workiva.net/organizations/Workiva/batch-changes/web_skin_dart_all_import)

### DIFF
--- a/test/executables/mui_migration_test.dart
+++ b/test/executables/mui_migration_test.dart
@@ -129,7 +129,7 @@ dependencies:'''),
           script: muiCodemodScript,
           input: inputFiles(additionalFilesInLib: [
             d.file('usage2.dart', /*language=dart*/ '''
-              import 'package:web_skin_dart/component2/all.dart';
+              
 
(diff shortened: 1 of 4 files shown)


aviary-wf:
## Security Insights

## Action Items
---

PR #280, opened 2024-03-29
Remove web_skin_dart ui_component.dart and ui_core.dart usages; use granular imports

sourcegraph-wk:
web_skin_dart has exports for individual components and an "ui_component" export that has all components.  If everywhere uses the individual component exports, then the compiler will only have to consider the component code it needs, resulting in faster build times.  But, this only works if everywhere is using the individual component exports.
 
 This batch:
 1. removes the ui_components.dart import. (And the deprecated ui_core.dart, opportunistically)
 2. puts in imports of every one of the individual components
 3. Uses `dart fix` to remove unused imports, unnecessary imports and duplicate_imports, which will get rid of components you don't need.
   * This may cause some side effects in other files if there are imports that are caught by one of the fixers.
   * There are some redundant imports that are not caught by the fixers, so you may need to manually remove them.
   * Namespaced imports will need manual repair
 
 Once CI is green, you can review and merge it.
 
[_Created by Sourcegraph batch change `Workiva/web_skin_dart_UI-component_import`._](https://workiva.sourcegraphcloud.com/organizations/Workiva/batch-changes/web_skin_dart_UI-component_import)

### DIFF
--- a/test/mui_suggestors/components/shared.dart
+++ b/test/mui_suggestors/components/shared.dart
@@ -18,7 +18,7 @@ String withOverReactAndWsdImports(String source) => /*language=dart*/ '''
     import 'package:over_react/over_react.dart';
     import 'package:web_skin_dart/component2/all.dart';
     import 'package:web_skin_dart/component2/all.dart' as wsd_v2;
-    import 'package:web_skin_dart/ui_components.dart' as wsd_v1;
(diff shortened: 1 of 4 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

PR #285, opened 2024-04-10
FED-2444 Null safety preparation for conditional prop function calls

aaronlademann-wf:
## Motivation
Functions passed in as prop values are often null checked before being called using patterns that do not migrate nicely to null-safety - requiring the use of the `!` assertion like so:

```dart
if (props.onClick != null) {
  props.onClick!(event);
}
// or
if (props.onClick != null) props.onClick!(event);
```

## Changes
1. Add a suggestor that removes the separate conditional null check and instead adds a null-aware `?` operator
1. Add the suggestor to the existing `null_safety_prep` executable.
1. Write tests.

### What it does
**Before**
```dart
if (props.onClick != null) {
  props.onClick(event);
}
// or
if (props.onClick != null) props.onClick(event);
```

**After**
```dart
props.onClick?.call(event);
```


aviary-wf:
## Security Insights

## Action Items
---

test/dart3_suggestors/null_safety_prep/fn_prop_null_aware_call_suggestor_test.dart
```diff
@@ -0,0 +1,225 @@
...
+                  }, [props.onClick]);
+                  
+                  return (Dom.button()..onClick = handleClick)();
+                },
+                UiFactoryConfig(displayName: 'Foo'),
+              );
+            '''),
+        );
+      });
+    });
```
greglittlefield-wf, line 134: Could we also add test cases that
- if-checks that don't involve props aren't updated (and the codemod skips over them without throwing)
- then statements that `return` something aren't updated
- then statements that do something besides call the prop aren't updated

greglittlefield-wf (changes requested):
Looking good, left some comments on cases to handle and tests

aaronlademann-wf:
All feedback addressed!

lib/src/dart3_suggestors/null_safety_prep/fn_prop_null_aware_call_suggestor.dart
```diff
@@ -0,0 +1,178 @@
...
+    final propFunctionBeingNullChecked =
+        _getPropFunctionBeingNullChecked(condition);
+    final ifStatement = parent;
+    if (ifStatement.elseStatement != null) return null;
+    if (ifStatement.parent
+            ?.thisOrAncestorOfType<IfStatement>()
+            ?.elseStatement !=
+        null) {
+      // There is an else-if statement present
+      return null;
```
greglittlefield-wf, line 115: I think this check isn't as specific as it should be, and due to the `thisOrAncestorOfType`, might cause cases to get skipped if they're nested. For example, I'd expect this to get skipped with the current logic
```dart
if (condition) {
  if (props.onClick != null) props.onClick(e)
} else {}
```

Could we just check the parent directly instead?
```suggestion
    if (ifStatement.parent.tryCast<IfStatement>()?.elseStatement ==
        ifStatement) {
      // ifStatement is an else-if
      return null;
    }
```

greglittlefield-wf (changes requested):
Thanks for addressing that feedback! Just one more small comment, otherwise LGTM; I also tested it locally and it works great!

greglittlefield-wf (approved):
+10

greglittlefield-wf (approved):
+10

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #322, opened 2025-08-07
FED-4061 Use gha-dart-oss

sydneyjodon-wk:
We should set up gha-dart-oss so we can:
- take advantage of shared workflows
- use the publish workflow to auto-publish on merge 🎉 

## QA instructions
- [ ] CI passes
- [ ] All steps previously run in CI are still running even if some of them are in the shared jobs now
- [ ] When this merges, follow along to make sure the package actually publishes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,13 +11,13 @@ on:
 
 jobs:
   checks:
-    uses: Workiva/gha-dart-oss/.github/workflows/checks.yaml@v0.1.11
+    uses: Workiva/gha-dart-oss/.github/workflows/checks.yaml@v0.1.12
     with:
       additional-checks: |
         no_entrypoint_imports
 
   build:
-    uses: Workiva/gha-dart-oss/.github/workflows/build.yaml@v0.1.11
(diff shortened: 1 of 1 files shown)


sydneyjodon-wk:
The failing additional-checks will be fixed once https://github.com/Workiva/gha-dart-oss/pull/42 merges

.github/workflows/ci.yml
```diff
@@ -10,7 +10,16 @@ on:
...
+  checks:
+    uses: Workiva/gha-dart-oss/.github/workflows/checks.yaml@v0.1.11
+    with:
+      additional-checks: |
+        no_entrypoint_imports
+
+  build:
+    uses: Workiva/gha-dart-oss/.github/workflows/build.yaml@v0.1.11
+
+  test:
```
sydneyjodon-wk, line 22: since this was renamed, I'll update Rosie before this merges to look for the renamed check

greglittlefield-wf: Updated! Unfortunately had to add each-sub job

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

